PR #32, opened 2016-08-08
Add Digilent jury member

wallento:


olofk:
Do we need a pull request for this? I'd say just push it

wallento:
yes, I do them to trigger @imphil to deploy it :)

olofk:
Ah ok. I guess I should had done the same then for my additions to the planet. I'll remember to do that from now on

PR #74, opened 2016-10-23
Revise routing scheme

imphil:
- Motivation: avoid namespace collisions between pages which are
  essentially named by a user (such as project and user profile pages),
  and pages which are defined by us.
- Motivation 2: logically structure content to make URLs
  self-explaining.
- The new routing scheme is described in routing.yml
- In addition to changing the routing scheme, we also split the
  controller into more logically separate parts, as started already for
  the Organization controller.


eliaskousk:
@imphil I reviewed all the changes and they are good, I couldn't find anything broken besides the search controller having a wrong name. It's now `searchProjectAction` instead of `searchAction` as the related controller route expects it. Since this is now a general search page I suggest that we change the name of the controller to `searchAction` instead of keeping the `searchProjectAction` and changing the route.

Now to the more important discussion regarding the routes. Although I like all of the new ones and what you improved, I have a suggestion to make: I think we should completely remove the `/p` prefix from the project view etc namespace.

This will allow us to have near identical routes with Github and make them easier for users to write and remember. Since you correctly removed the `/p` prefix from the "app level" routing there is now no need to keep it inside the bundle level routing at all. We can just move this "open" namespace at the end of the file and there won't be any collisions with the rest of the routes since they have a higher priority by Symfony.

See here my changes on the bundle routes:

https://gist.github.com/eliaskousk/06670dfd1e0d125f628539c2ea4dda95

In order for everything above to work with the rest of the bundles we need a minor change in the static routes (site pages) that now need to be placed in their own `static` namespace (or `s` if you prefer to make it shorter) and moved before the last `/` namespace which is open to the users:

https://gist.github.com/eliaskousk/38e374b3b2475980c0d35a131e28d5f1

I believe its better to sacrifice the static routes and make them a little longer in order to have the gain of users being able to have their LibreCores projects in the top-level namespace.

Let me know of what you think of the above and if you agree I can do a commit on your PR branch with them. If you don't agree I believe its also fine as it is and you can proceed with the PR merge. (after fixing the search controller name!)

imphil:
@eliaskousk thanks for having a look at it. I like your idea, given that the most commonly "manually typed" URLs will be the ones in the project repo. So let's go ahead with moving the site pages into `/static/`. By doing that we'll only need two more adjustments: disallow "static" as project name in the UserProjectNameValidator, and find a new name for the start page of the project repository (I guess just using `/projects` will work fine, but I'm open to whatever makes sense)

Just go ahead and commit it right into this branch, that's why I put it in this repo for both of us to have access.

imphil:
Maybe one additional comment: You write "We can just move this "open" namespace at the end of the file and there won't be any collisions with the rest of the routes since they have a higher priority by Symfony." That's correct, but only half the story. In order to avoid users creating usernames or organization names which nobody can access, i.e. names which clash with system-defined ones, all such names must be explicitly disallowed by the user/org name validator. If we add new routes in the future, we must ensure that we don't clash with already existing user/org names. 

To give us some breathing room for future extensions, we currently limit the minimum name to 4 characters, i.e. we can use all routes with up to 3 characters without risking a name clash with an existing user or org name. I guess that's sufficient for future-compat.

eliaskousk:
Yes you are right, we must forbid any names that clash with system-defined ones. I reviewed the userorg name validator and it already contained some of those. I removed some unneeded and added all the needed ones (two commits, forgot some on the first, sorry!).

I also added a controller that gives us the ability to redirect URLs with trailing slashes into URLs without it. E.g `/search/` will redirect to `/search` etc. I believe this is necessary as to not error out with a seemingly correct URL like `/username/projectname/` and thus confuse the user.

site/src/Librecores/ProjectRepoBundle/Validator/Constraints/UserOrgNameValidator.php
```diff
@@ -37,7 +37,8 @@ class UserOrgNameValidator extends ConstraintValidator
      *
      * @var string[]
      */
-    const RESERVED_NAMES = [ 'org', 'orgs', 'project', 'projects', 'search', 'unassigned' ];
+    const RESERVED_NAMES = [ 'org', 'orgs', 'planet', 'project', 'projects',
+                             'search', 'static', 'unassigned', 'user' ];
```
imphil: Since this lookup is not in any hot path, can we, just to be safe and in addition to the list of reserved names, do a `doesThisRouteExist($userOrgName)` lookup (however that is actually called in Symfony)?

eliaskousk: Yes we can call $router->match() on the route that get created with `/userName` and see if it matches. E.g in validate() of the user/org name validator use the following:

```php
         if ($this->userOrOrgNameExists($value)) {
            $this->context
                ->buildViolation($constraint->messageUniqueName)
                ->setParameter('%string%', $value)
                ->addViolation();
        }
        else if ($this->userOrOrgReserved($value)) {
            $this->context
                ->buildViolation($constraint->messageReservedName)
                ->setParameter('%string%', $value)
                ->addViolation();
        }
```

where `userOrOrgReserved()` is defined as:

```php
    /**
     * Find out if a given top-level URL value is reserved
     * @param $value
     * @return bool
     */

    private function userOrOrgReserved($value)
    {
        return (in_array($value, self::RESERVED_NAMES) ||
                $this->router->match('/' . $value)['_route']);
    }
```

As we stand now the match() part of the above function is already covered by userOrgNameExists(). However I agree that this is more safe in case we add in the future new routes and forget to update the reserved names of the validator.

imphil:
Thanks Elias, I've pushed it to master. Let's continue with polishing the organization support then.

Issue #80, opened 2016-12-15
Show project activity metrics


asb:
I thought the Visual Studio Code plugin index had a [nice idea](https://marketplace.visualstudio.com/items?itemName=vscodevim.vim). For projects on GitHub, list activity metrics, e.g.:
9 Pull Requests
97 Open Issues
Last commit: 4 days ago

oleg-nenashev:
It would be very useful. The only problem is that polling much things from GitHub may finally lead to the violation of the REST API limit. I see two possible ways:
* Cache the data in LibreCores and update on the daily basis
* Just let Web Browsers retrieve the data on their own (requires some JavaScript coding)

On the other hand, it can be easily done just by embedding GitHub shields (also safe for API calls limit IIRC). Here is a good list of available Shields (including GitHub ones): http://shields.io/

asb:
And you could just have a server-side process retrieve the .svg from shields.io and extract the text for formatting in your preferred form. REST API problem solved, right? :)

asb:
Actually they support .json output too (https://img.shields.io/github/issues-closed/badges/shields.json).

berndca:
The [github REST API rate limit for authenticated requests](https://developer.github.com/v3/#rate-limiting) is 5000 requests per hour. 

I had no problems using the API to extract repo info about  [open source HDL repos](https://berndca.github.io/discover_fossi/).  @oleg-nenashev  Is there a specific issue you are referring to?

I would create an AWS lamdba CRON job to extract all required info to a single json file perhaps once a day which I don't think qualifies for **aggressive polling**. I'd be happy to help setting things up.

oleg-nenashev:
@berndca 

> I had no problems using the API to extract repo info about open source HDL repos. @oleg-nenashev Is there a specific issue you are referring to?

Well, we have issues with the GitHub API limit for authenticated users. E.g. in Jenkins project there were a couple of recent Infra issues due to it.

oleg-nenashev:
Created #83

eliaskousk:
Updated #83

eliaskousk:
We can also use the [Github API client library](https://github.com/KnpLabs/php-github-api) which seems great for our purposes and should be very easy to use inside our project.***

Please see their [doc](https://github.com/KnpLabs/php-github-api/tree/1.7/doc) folder for a lot of examples on various data that we could get from each hosted project. We can identify a set of things that are useful for LibreCores and fetch them using this library.

[...]

PR #83, opened 2016-12-18
Project page: Add badges for projects being hosted on GitHub

oleg-nenashev:
Addresses #80

The PR provides the following badges:

* Number of stars (if SourceRepo URL is GitHub)
* Number of open pull requests (if SourceRepo URL is GitHub)
* Number of open issues (if Issue Tracker URL is GitHub)

**Disclaimer**: The code below likely contains my first code lines in PHP ever. Will appreciate any reviews regarding making the code compliant with project standards.

Example:
screen shot 2016-12-18 at 18 01 45

CC @asb and @berndca

### DIFF
--- a/site/src/Librecores/ProjectRepoBundle/Entity/Project.php
+++ b/site/src/Librecores/ProjectRepoBundle/Entity/Project.php
@@ -4,6 +4,7 @@ namespace Librecores\ProjectRepoBundle\Entity;
 use Doctrine\ORM\Mapping as ORM;
 use Symfony\Component\Validator\Constraints as Assert;
 use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
+use Librecores\ProjectRepoBundle\Util\GitHubUtil;
 
 /**
  * A project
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
As we discussed in #80, this approach makes the user web browser to send GitHub requests and to consume it's API limit. It's not the best solution. Likely a kind of the background crawler is required.

eliaskousk:
Thank you @oleg-nenashev, that was a great first try on PHP. I improved the PR by adding Github util as a service in the container which is the recommended way of adding such things in Symfony. I also refactored the code and in general made everything Symfony friendly.

Looking at the changes you can see how a service is defined in YAML and retrieved from a controller and how to pass it static arguments (e.g the github.com and shield.io URLs). Dynamic arguments like things that a controller generates are best set by setter functions and not in the service's constructor (e.g the projects source repo and issue tracker url). Setting such parameters for end user code is a little overkill (github.com would never change!) but I thought it was a good example on parameter passing.

I don't have access and couldn't push to @oleg-nenashev's  fork in order for it to show here and so I pushed the commit to a separate branch. You can find it on 22f9e34.

oleg-nenashev:
@eliaskousk Just create a new PR for your fork. Thanks for adjusting it

eliaskousk:
Closing this

PR #84, opened 2016-12-19
Project page: Add badges for projects being hosted on GitHub (NEW)

eliaskousk:
Original PR #83 from @oleg-nenashev:

Addresses #80

The PR provides the following badges:

* Number of stars (if SourceRepo URL is GitHub)
* Number of open pull requests (if SourceRepo URL is GitHub)
* Number of open issues (if Issue Tracker URL is GitHub)

**Disclaimer**: The code below likely contains my first code lines in PHP ever. Will appreciate any reviews regarding making the code compliant with project standards.

Example:
screen shot 2016-12-18 at 18 01 45

CC @asb and @berndca

### DIFF
--- a/site/src/Librecores/ProjectRepoBundle/Util/GitHubUtil.php
+++ b/site/src/Librecores/ProjectRepoBundle/Util/GitHubUtil.php
@@ -3,47 +3,134 @@
 namespace Librecores\ProjectRepoBundle\Util;
 
 /**
- * Helper class, which provides methods for operating with GitHub projects.
+ * Helper utility service, which provides methods for operating with GitHub projects.
  */
-class GitHubUtil {
-  
-  private $url;
-  
-  public function __construct($url) {
-    $this->url = $url;
-  }
-  
-  public function isOnGitHub()
-  {
-    if (strpos($this->url, 'github') !== false) {
-      return true;
+
+class GitHubUtil
+{
+
+    private $githubUrl;
+    private $shieldsUrl;
+    private $shieldsExt;
+    private $sourceRepoUrl;
+    private $issueTrackerUrl;
+
+    public function __construct($githubUrl, $shieldsUrl, $shieldsExt)
+    {
+        $this->githubUrl       = $githubUrl;
+        $this->shieldsUrl      = $shieldsUrl;
+        $this->shieldsExt      = $shieldsExt;
+        $this->sourceRepoUrl   = null;
+        $this->issueTrackerUrl = null;
     }
-    return false;
-  }
-  
-  public function getStarsBadgeImage()
-  {
-    return 'https://img.shields.io/github/stars/' . $this->getGitHubPath($this->url) . '.svg';
-  }
-  
-  public function getIssuesBadgeImage()
-  {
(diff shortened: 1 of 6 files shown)


eliaskousk:
Is there an easy way to move (rebase?) the 2 commits that fix the validation bug out of this branch / PR? because the original PR from @oleg-nenashev started from an earlier commit on master (without them) but I already did and thus this situation. I am always afraid of rebasing on repos that I don't fully control so if  anyone wants to do it please feel free :-P

site/src/Librecores/ProjectRepoBundle/Util/GitHubUtil.php
```diff
@@ -0,0 +1,136 @@
...
+    private $sourceRepoUrl;
+    private $issueTrackerUrl;
+
+    public function __construct($githubUrl, $shieldsUrl, $shieldsExt)
+    {
+        $this->githubUrl       = $githubUrl;
+        $this->shieldsUrl      = $shieldsUrl;
+        $this->shieldsExt      = $shieldsExt;
+        $this->sourceRepoUrl   = null;
+        $this->issueTrackerUrl = null;
```
oleg-nenashev: IMHO this is a bad approach. Once we add more GitHub descriptors (e.g. for releases or Wiki), we will need to add more URLs to this class. It makes the logic a bit less flexible then it was in the original approach, when the class was being constructed once for a single URL

oleg-nenashev:
@eliaskousk Well, it's hard to review the stuff now. Ideally tou should have checked out my branch to a new local one and then continue working in it.

eliaskousk:
@oleg-nenashev That's what (I think!) I did, but maybe when pulling, auto rebasing replayed my master commits on top of yours. I will try to make sure this doesn't happen next time.

site/src/Librecores/ProjectRepoBundle/Util/GitHubUtil.php
oleg-nenashev: > I kept the original separation of sourceRepo and issueTracker URLs because I liked the original intention of allowing a project to have the source repo and issue tracker separate.

Yes, this is must-have. Many big projects use external bugtrackers like JIRA

> If it starts getting complicated we could wrap it in a factory service but I don't think there are so many Github URLs to justify this. Even adding a couple more can be done quickly as it is

OK. What we may really need:
* Releases support (with proper badges, etc.)
* Maybe "Maintainers" link pointing to a GitHub team

eliaskousk: Yes these are nice to have, we should add them, at least the links to them (I suppose you mean /releases, /graph). For more info let's bite the bullet and use their api. I found a cool client library, more info on #80.

imphil:
Thanks @eliaskousk  and @oleg-nenashev for the work on this PR. 

I have thought about all this GitHub API stuff, also in conjunction with the crawler thing to update the project data, a bit during the last days. I'll get back soon once I've finished thinking, and make sure we don't have a (mental) merge conflict here. So please leave this open for now, it's not forgotten.

imphil:
Closing this now we have the same metrics as presented in the badges through the GH api. Please reopen if I missed something.

PR #87, opened 2017-01-03
Update RSS for Nyuzi blog

olofk:


imphil:
thanks olof, will go to production shortly

PR #88, opened 2017-01-09
Fix the FOS User Manager service

eliaskousk:
Its constructor signature has changed on the latest development version so our derived user manager service needs correct injection of needed services.

### DIFF
--- a/site/src/Librecores/ProjectRepoBundle/Resources/config/services.yml
+++ b/site/src/Librecores/ProjectRepoBundle/Resources/config/services.yml
@@ -22,7 +22,7 @@ services:
 
     librecores_user_manager:
         class: Librecores\ProjectRepoBundle\Doctrine\UserManager
-        arguments: [@security.encoder_factory, @fos_user.util.username_canonicalizer, @fos_user.util.email_canonicalizer, @fos_user.object_manager, %fos_user.model.user.class%]
+        arguments: ['@fos_user.util.password_updater', '@fos_user.util.canonical_fields_updater', '@fos_user.object_manager', '%fos_user.model.user.class%']
 
     librecores_user_provider:
         class: "%librecores_user_provider.class%"


imphil:
Looks like this change depends on an composer update, which in turn requires more changes to adjust to the new FOSUserManager version.

With the php71 branch, which includes an updated composer setup, I get

~~~
Attempted to call an undefined method named "canonicalizeUsername" of class "Librecores\ProjectRepoBundle\Doctrine\UserManager".
500 Internal Server Error - UndefinedMethodException 
~~~

It doesn't look like this is fixed in eliask/org.polish yet; I'll look into it tonight (if nobody beats me to it) :)

eliaskousk:
Yes, sorry I forgot to mention the need for composer update, I rarely use the locked versions on dev machines.

Now I looked around the parent fos user manager (latest version) and indeed it is missing the `canonicalizeUsername` function. It is moved inside the new `canonicalFieldsUpdater` injected class / service of the user manager. So we need a change in our derived user manager's `findUserByUsernameOrEmail2` function so that it calls `$this->canonicalFieldsUpdater->canonicalizeUsername($username)` instead of `$this->canonicalizeEmail($email)`.

The strange thing is that I don't get the error anywhere and I can't find where `findUserByUsernameOrEmail2` is used so that I can test the change! Let me know if the above fixes it or where `findUserByUsernameOrEmail2` is called so that I can test the change myself :P

eliaskousk:
OK I found the usage, its in the oauth system and I only test with simple logins not Github etc. Should I do the commit or will you test first and do it yourself?

imphil:
yes, it's in the OAuth login. I'll test and push it, thanks.

imphil:
OK, thanks for the hint, pushed fbe785ad5695865cd03301b2588aa3e20e357ec3 together with the composer update and PHP 7.1.

PR #89, opened 2017-01-09
Remove PHP 7 PPA (ondrej/php)

eliaskousk:
This PPA breaks PHP 7 extensions. Using the PPA the system contains both PHP 7.0 and PHP 7.1 extensions and so the PHP 7.1 ones don't get loaded on the installed PHP 7.0 interpreter (e.g the XML / SimpleXML extension is missing on the FPM but not on the CLI).

You can reproduce the problem just by destroying and re-creating your local VM (or provisioning it, I haven't tried it though).

I suppose we can keep the PPA and move to PHP 7.1 which is already released and seems pretty stable. If we stay on PHP 7.0 though we must remove the PPA.

### DIFF
--- a/ansible/roles/web/tasks/php.yml
+++ b/ansible/roles/web/tasks/php.yml
@@ -1,7 +1,10 @@
 ---
 # Install PHP for the web server
-- name: Add PHP ppa repository
-  apt_repository: repo='ppa:ondrej/php'
+
+#
+#- name: Add PHP ppa repository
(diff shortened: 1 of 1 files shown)


imphil:
@eliaskousk thanks for this PR. I've solved the problem a little bit differently, leaving the PPA in place and explicitly specifying the PHP version explicitly. I've tried updating straight to PHP 7.1, however, there are still some migration issues to be solved first (I'll work on that in the php71 branch)

PR #101, opened 2017-01-29
Organizations

eliaskousk:
The work on organizations is mostly complete and supports the following:

- Create an org
- Edit an org
- Remove an org
- Add a project under an org (at last!!!)
- Edit a project under an org
- Join an org that a user is not already a member or denied previously
- Any member of the org can approve or deny a request to join the org
- View an org and it's projects, members, settings, join requests, etc.
- View the orgs a user belongs or requested to join in his profile
- View the orgs a user belongs or requested to join in a separate page for this purpose

Each organization entity (`Organization`) is tied to a user (`User`) via a mapping called `OrganizationMembers`. This mapping besides the foreign keys holds permissions and other meta-data that we want to store there. Currently besides permissions only timestamps are stored but this may change in the future.

Roles / permissions are quite flexible in this setup and we currently support the following roles:

* **REQUEST** - The user has requested to join the org

* **DENY** - The user was denied joining the org

* **SUPPORT** - The user supports / follows the org but he is not a member. Currently not used, but I thought we might want to allow user to follow projects, news etc. on orgs without officially being a member, like a user can 'star' projects. Ideas on this are welcome!

* **MEMBER** - The user is a member of the org and has full control of it, e.g approving new members, removing other members, editing settings etc.

* **ADMIN**  - The user is an admin of the org. Currently set only for the user who created the org. I added this role in-case we want to allow a member super-special privileges like deleting the org and all projects.

Each role can be extended in the future with fine-grained permissions that are stored in a separate entity e.g `OrganizationPermissions` and associated from the current `permissions` property, thus requiring minimal changes. For now though I think the above 5 roles are more than enough.

Lastly the `Organization` entity still contains the original `creator` of the org as a property. We don't use this anywhere and it's only there for logging / history purposes. This creator as of the latest commits doesn't have any special power that the other members do not. We can remove this property without any consequences but I think we should keep it as it does no harm.

More discussion on the rationale of the above decisions can be found in the main issue #7.


site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
```diff
@@ -106,10 +146,12 @@ public function viewAction(Organization $organization)
      */
     public function settingsAction(Request $request, Organization $organization)
     {
-        if ($this->getUser() != $organization->getOwner())
-            throw $this->createAccessDeniedException("You don't own this organization in order to make changes");
+        if (!$this->userIsMember($organization)) {
+            throw $this->createAccessDeniedException(
+                "You aren't a member of the organization in order to make changes.");
```
imphil: "You need to be member of this organization in order to make changes."

site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
```diff
@@ -121,198 +163,247 @@ public function settingsAction(Request $request, Organization $organization)
...
-    public function searchAction()
-    {
-        // Search for an organization
-
-        return $this->render('LibrecoresProjectRepoBundle:Organization:search.html.twig');
-    }
-
-    /**
-     * Request to join an organization
+     * Leave from an organization
```
imphil: Nit: "Leave an organization"

site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
```diff
@@ -121,198 +163,247 @@ public function settingsAction(Request $request, Organization $organization)
...
 
-        // Join the organization
-
         $user = $this->getUser();
-        $o->addRequest($user);
+
+        // Remove the organization membership
+        $member = $this->getDoctrine()
+                       ->getRepository('LibrecoresProjectRepoBundle:OrganizationMember')
+                       ->findOneBy(['organization' => $o,
```
imphil: so nice to finally use the new array syntax! :-)

site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php
```diff
@@ -26,45 +25,60 @@ public function newAction(Request $request)
...
+
+                $p->setParentUser($user);
+
+            } else if ($formParentType === 'o') {
+                $organization = $this->getDoctrine()
+                                     ->getRepository('LibrecoresProjectRepoBundle:Organization')
+                                     ->findOneByName($formParentName);
+
+                $p->setParentOrganization($organization);
+            }
```
imphil: All of this is still hacky as hell, but let's leave it for now. I'll open a follow-up bug to get this fixed properly (doesn't seem to be straightforward though).

site/src/Librecores/ProjectRepoBundle/Entity/Organization.php
```diff
@@ -59,37 +60,43 @@ class Organization
...
+     * @ORM\JoinColumn(name="creatorId", referencedColumnName="id")
+     */
+    private $creator;
+
+    // metadata
+    /**
+     * When was this organization created?
+     *
+     * @var \DateTime
+     * @ORM\Column(type="datetime", nullable=true)
```
imphil: Shouldn't this always be present? nullable = false then

site/src/Librecores/ProjectRepoBundle/Entity/Organization.php
```diff
@@ -59,37 +60,43 @@ class Organization
...
+     */
+    protected $createdAt;
+
+    /**
+     * When was this organization last updated?
+     *
+     * @var \DateTime
+     * @ORM\Column(type="datetime", nullable=true)
```
imphil: It's much easier in frontend code not to check for updatedAt === null, but to always get a DateTime object. Since below in PrePersist() it's set anyways on every storing of the data, just make this nullable = false as well?

site/src/Librecores/ProjectRepoBundle/Entity/OrganizationMember.php
```diff
@@ -0,0 +1,204 @@
...
+ * @ORM\Entity
+ * @ORM\HasLifecycleCallbacks
+ */
+class OrganizationMember
+{
+    const PERMISSIONS_REQUEST = 'REQUEST';
+    const PERMISSIONS_DENY    = 'DENY';
+    const PERMISSIONS_SUPPORT = 'SUPPORT';
+    const PERMISSIONS_MEMBER  = 'MEMBER';
+    const PERMISSIONS_ADMIN   = 'ADMIN';
```
imphil: nit: these constants just name a single permission, e.g. PERMISSION_REQUEST. Feel free to ignore if you disagree.

site/src/Librecores/ProjectRepoBundle/Entity/OrganizationMember.php
```diff
@@ -0,0 +1,204 @@
...
+        return $this->organization;
+    }
+
+    /**
+     * Set permissions
+     *
+     * @param string $permissions One of the self::PERMISSIONS_* constants
+     * @throws \InvalidArgumentException
+     */
+    public function setPermissions($permissions)
```
imphil: setPermissions() suggests that you set an array of permissions, but in fact it's only one. So rename to setPermission()?

site/src/Librecores/ProjectRepoBundle/Entity/OrganizationMember.php
```diff
@@ -0,0 +1,204 @@
...
+        $this->permissions = $permissions;
+    }
+
+    /**
+     * Get permissions
+     *
+     * @return string
+     */
+    public function getPermissions()
+    {
```
imphil: same as for setPermissions()

site/src/Librecores/ProjectRepoBundle/Entity/OrganizationRepository.php
```diff
@@ -12,42 +12,23 @@
...
-            ->getQuery()
-            ->getResult();
-    }
-
-    /**
      * Find all the organizations that a user is a member of.
      *
      * @param User $user
-     * @return ArrayCollection of Organizations
+     * @return array of Organizations
```
imphil: probably nowadays the "correct" annotation would be `@return Organization[]`

site/src/Librecores/ProjectRepoBundle/Validator/Constraints/UserOrgNameValidator.php
```diff
@@ -76,7 +76,7 @@ class UserOrgNameValidator extends ConstraintValidator
      *
      * @var bool
      */
-    const CHECK_UNIQUE_ORG  = true;
+    const CHECK_UNIQUE_ORG  = false;
```
imphil: Doesn't this allow users to be created with the same name as an organization, and the other way around? Right now this case will be caught (rather by luck) in `UserOrgNameValidator::userOrOrgReserved()`, but let's not depend on that. I'm proposing to remove both `CHECK_*` constants and unconditionally do the check in `userOrOrgNameExists()`. It's not in a hot path anyways, so the overhead of double-checking should be acceptable.

sql/dev_fixtures_site.sql
```diff
@@ -1,8 +1,8 @@
 -- sample data for development of the LibreCores web site
 
-INSERT INTO `Organization` (`id`, `name`, `displayName`, `description`, `ownerId`) VALUES
-(1, 'unassigned', 'Unassigned', 'Unassigned description', NULL),
-(2, 'openrisc', 'OpenRISC', 'OpenRISC Description', NULL);
+INSERT INTO `Organization` (`id`, `name`, `displayName`, `description`, `creatorId`, `createdAt`, `updatedAt`) VALUES
+(1, 'unassigned', 'Unassigned', 'Unassigned description', NULL, NULL, NULL),
+(2, 'openrisc', 'OpenRISC', 'OpenRISC Description', NULL, NULL, NULL);
```
imphil: If you make this not-null above, this also nees to change (until we have proper fixtures some day) :)

site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
```diff
@@ -121,198 +163,247 @@ public function settingsAction(Request $request, Organization $organization)
...
-
-    /**
-     * Request to join an organization
+     * Leave from an organization
      *
      * @param string $organizationName
      * @return Response
      */
-    public function joinAction($organizationName)
+    public function leaveAction($organizationName)
```
imphil: Currently it's possible for the admin/creator to leave the organization, effectively causing the organization to be a zombie. Can we prevent that here (or somewhere else)?

site/src/Librecores/ProjectRepoBundle/Entity/Organization.php
wallento: I want to second that. Currently working on "recent activity" we want that information!

eliaskousk:
Thanks for the review! I'll reply to your comments inline. I'll also fix the double linking of the Profile when I rebase later today.

site/app/Resources/views/base.html.twig
```diff
@@ -58,18 +58,35 @@
           {% if is_granted("IS_AUTHENTICATED_REMEMBERED") %}
           <li class="dropdown">
             <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
-              {{ 'layout.logged_in_as'|trans({'%username%': app.user.username}, 'FOSUserBundle') }}
+              {{ 'layout.logged_in_as'|trans({'%username%': app.user.usernameCanonical}, 'FOSUserBundle') }}
```
eliaskousk: You are right, this was a force of habit of mostly using canonical usernames and templates are not the right place for it.

site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
eliaskousk: Ok

site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
eliaskousk: Ok

site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
eliaskousk: Should we prevent leaving or allow it and set the organization to a default e.g 'librecores' creator? I don't have a preference, your call!

site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php
eliaskousk: Ok

site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
imphil: let's just prevent leaving for now, and add a TODO mark to implement this together with deleting of organizations (maybe you can open an enhancement issue for this so it doesn't get lost?)

site/src/Librecores/ProjectRepoBundle/Entity/Organization.php
eliaskousk: It was false until yesterday and I agree with what you say. However I was forced to change it when I encountered an exception from mysql (can't recall the exact situation) that I didn't have time to resolve. I'll try to investigate further and see how it goes so that we can set it back to false.

site/src/Librecores/ProjectRepoBundle/Entity/Organization.php
eliaskousk: Agreed, see above.

site/src/Librecores/ProjectRepoBundle/Entity/Organization.php
imphil: maybe coming from the imported sql we use as pseudo-fixtures?

site/src/Librecores/ProjectRepoBundle/Entity/OrganizationMember.php
eliaskousk: I mostly agree and was set to remove the 'S' but I decided not to since these might actually represent a set of permissions not only a single fine-grained one. Currently I think they are very simple and justify either meaning so I can't decide. Maybe rename to ROLE_ but this mentally classes with the symfony security roles. Uhh... why everything must be so complicated! Let's stick with the basic rule that until something changes what it does it's best to name it accordingly to what it does NOW. So I agree let's lose the 'S'.

site/src/Librecores/ProjectRepoBundle/Entity/OrganizationMember.php
eliaskousk: Agreed, see above.

site/src/Librecores/ProjectRepoBundle/Entity/OrganizationMember.php
eliaskousk: Agreed, see above.

site/src/Librecores/ProjectRepoBundle/Entity/OrganizationRepository.php
eliaskousk: Agreed, I made the change to make a code check warning pass but what you mention is the better fix which keeps the type and I'm positive it will pass the check as well.

site/src/Librecores/ProjectRepoBundle/Validator/Constraints/UserOrgNameValidator.php
eliaskousk: The unique username check results to the generation of two constraint validation warnings on the user's form and that's why it was originally removed from here. Thus we must keep either the unique check on the entity (doctrine annotation) or here but not both. Maybe there is another solution that keeps both but only shows one on the form?, I haven't searched for that though, maybe it's possible, in relation to constraint validation groups, see below.

The unique org name check was removed because keeping it makes it trigger everytime you edit the organization settings and just keep / submit the same name (99.9% of cases). I've tried removing the field from the form as to disallow the editing of the name by users altogether, like we do on projects, but it keeps triggering. We can solve this only by having constraint validation groups that trigger the constraints on selected CRUD operations. Thus we should trigger certain constraints only on org creation but not on org updates etc. Maybe leave it as is now and make this in a future change-set that is aimed at having more complete constraints?

sql/dev_fixtures_site.sql
eliaskousk: Agreed, we must place a proper date/time here if these become not null. And proper fixtures are on the way (TM)! :)

site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
eliaskousk: Yes! It really feels more modern and aligned with other languages that have drooooling features (unfortunately on PHP we only get a handful of them).

site/src/Librecores/ProjectRepoBundle/Validator/Constraints/UserOrgNameValidator.php
imphil: > The unique username check results to the generation of two constraint validation warnings on the user's form and that's why it was originally removed from here.

Yeah, you'll get two validation warnings if you create a user with the same name as an existing user. However, you'll get only one validation error if you create an user with the same name as an organization. However I'd consider that a minor annoyance, compared to us creating an org and a user with the same name :-)

I don't know if there's a better way of having a UNIQUE constraint across two columns on different tables; I didn't find one in Doctrine when I wrote that one. 

We should check if we can safely remove the unique validation constraint from User and Organization, while keeping the DB constraint (which is useful for index generation).



> The unique org name check was removed because keeping it makes it trigger everytime you edit the organization settings and just keep / submit the same name (99.9% of cases). I've tried removing the field from the form as to disallow the editing of the name by users altogether, like we do on projects, but it keeps triggering.

Yeah, validation groups should be the answer here. I don't know it off-head, so if you cannot figure it out let me know and I'll have a look.


> Maybe leave it as is now and make this in a future change-set that is aimed at having more complete constraints?

If we don't figure this out now we could run into users creating orgs which shadow existing users, and the other way around. I think this is something we don't want to happen and cannot really fix in the future without making people unhappy, so let's figure this out before it goes live.

site/src/Librecores/ProjectRepoBundle/Entity/Organization.php
eliaskousk: (If I recall correctly) I tried to solve it as you suggest but (again IIRC) it persisted. I'll see how it goes again and report back.

site/src/Librecores/ProjectRepoBundle/Validator/Constraints/UserOrgNameValidator.php
eliaskousk: I agree on all your points. With 'future changeset' I meant to do it before pushing this PR on production :P i.e make it this week. But since it is urgent to prevent this shadowing of users / orgs let's make the important changes on this PR and leave the rest for the more distant future after we have a better sense of our needs there. I'll try to experiment with the CVGs early this week and make the changes.

imphil: Sounds good. Just on a timing note, I'd love to get this integrated before FOSDEM next weekend, so maybe you can just update this PR with what you have tonight and then I can take it from there if I find more free time than you do. Possibly then Bernd can also have a look at the styling before next weekend, which would be fantastic.

eliaskousk: I knew you were going to FOSDEM and so I accelerated development (we have a great 'marketing' opportunity at FOSDEM!). I agree we should finish this PR sooner or later and Bernd hopefully will find the time to beautify it before next weekend.

site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php
eliaskousk: I agree.

eliaskousk:
@imphil Everything has been addressed as discussed and we can proceed. Some notes on the patches:

- The timestamps are back to non-nullable and it just works this time... I don't know what the hell happened last time that they didn't work as non-nullable, let's just forget about it :-P

- I also solved the validation constraint issue with what I think is the simplest and most elegant way for our situation. The root of all the issues was that we were using the same constraint validation class for two separate entities and both `select` orm queries on the validator run regardless of whether we were checking for unique usernames or unique org names. I eventually dismissed the constraint validation groups solution since it is an overkill for this situation, we really don't have much different needs for the different CRUD operations. Another solution I dismissed was to split the validation class in two new classes and thus having two constraints, one for usernames and one for orgs. I thought this was also overkill since these two classes would do 99.9% the same thing which means that your original design with one class is fine and we better keep it that way.

Finally I proceeded with a third solution that we didn't originally see. That is to use the `payload` property on the constraints. We now set it to `"user"` when checking usernames (against orgs) and to `"org"` when checking org names (against usernames). Inside the constraint validation class there is a check for the payload's value to decide what to check on the db. It doesn't need to check usernames against usernames since they are already forced unique (same for org names).

I did some testing and it works fine, with the added bonus of producing only one error message on form submissions. Yeay for payload! Yeay for Symfony! that one more time provides a good abstract solution for a variety of 'special' issues someone might have and not requiring constant fighting to solve problems.

imphil:
Thanks a lot Elias. Very nice work! I've given it a try and it looks very nice! I'll open some follow-up issues to improve on top of this.

wallento:
Hooorraayy!!!!

PR #102, opened 2017-01-29
Add activity to main page

wallento:
This is a first, straight forward approach, that should be extended later. But for the first it adds some dynamic content to the home site.

### DIFF
--- a/site/src/Librecores/ProjectRepoBundle/Entity/ProjectRepository.php
+++ b/site/src/Librecores/ProjectRepoBundle/Entity/ProjectRepository.php
@@ -58,4 +58,23 @@ class ProjectRepository extends EntityRepository
         }
         return $p->getResult();
     }
+
+    /**
+     * Find all projects ordered by recent activity (last modified)
+     *
+     * @param int $limit
+     * @return array of projects
+     */
+    public function findByRecentActivity($limit = null)
+    {
+		$p = $this->getEntityManager()
+			->createQuery(
+				'SELECT p '.
+				'FROM LibrecoresProjectRepoBundle:Project p '.
+				'ORDER BY p.dateLastModified DESC');
+		if ($limit != 0) {
+			$p->setFirstResult(0)->setMaxResults($limit);
+		}
+		return $p->getResult();
+    }
 }
--- a/site/src/Librecores/SiteBundle/Controller/DefaultController.php
+++ b/site/src/Librecores/SiteBundle/Controller/DefaultController.php
@@ -28,6 +28,9 @@ class DefaultController extends Controller
         // blog posts on planet
         $templateArgs['blogposts'] = $this->getBlogPosts();
 
+        $prjrepo = $this->getDoctrine()->getRepository('LibrecoresProjectRepoBundle:Project');
+		$templateArgs['activity'] = $prjrepo->findByRecentActivity(10);;
+
         return $this->render('LibrecoresSiteBundle:Default:home.html.twig',
             $templateArgs
         );
--- a/site/src/Librecores/SiteBundle/Resources/views/Default/home.html.twig
+++ b/site/src/Librecores/SiteBundle/Resources/views/Default/home.html.twig
@@ -94,65 +94,82 @@
   </div>
 </div>
 
-<!-- 3 columns with main messages -->
 <div class="container" style="padding-top: 30px;">
   <div class="row">
     <div class="col-md-6">
-      <h2>What is LibreCores?</h2>
-      <p>Digital devices (or commonly "chips") are the basic building
-      blocks of electronic systems. Such devices are for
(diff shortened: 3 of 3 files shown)


wallento:
Also see #103

wallento:
@berndca how could we improve the appearance?

berndca:
I like the idea of an activity protocol. However I believe it would be best to put the activity stream on another page for now to allow us to experiment with data sources and formats. This could be a page "Whats New at LibreCores?" perhaps under Community? Maybe we will move it or some version of it to the landing page in the future.

imphil:
What's the status on this one? Ready for review or will it be replaced by activity on a sub-page and can thus be closed for now?

wallento:
We should have both: This simple one for now and a followup development as suggested by @berndca.

site/src/Librecores/ProjectRepoBundle/Entity/ProjectRepository.php
```diff
@@ -58,4 +58,23 @@ public function findByFqnameFragment($fqnameFragment, $limit = null)
         }
         return $p->getResult();
     }
+
+    /**
+     * Find all projects ordered by recent activity (last modified)
+     *
+     * @param int $limit
+     * @return array of projects
```
imphil: `@return Project[]`

site/src/Librecores/ProjectRepoBundle/Entity/ProjectRepository.php
```diff
@@ -58,4 +58,23 @@ public function findByFqnameFragment($fqnameFragment, $limit = null)
         }
         return $p->getResult();
     }
+
+    /**
+     * Find all projects ordered by recent activity (last modified)
+     *
+     * @param int $limit
```
imphil: @param int|null $limit

site/src/Librecores/ProjectRepoBundle/Entity/ProjectRepository.php
```diff
@@ -58,4 +58,23 @@ public function findByFqnameFragment($fqnameFragment, $limit = null)
...
+			->createQuery(
+				'SELECT p '.
+				'FROM LibrecoresProjectRepoBundle:Project p '.
+				'ORDER BY p.dateLastModified DESC');
+		if ($limit != 0) {
+			$p->setFirstResult(0)->setMaxResults($limit);
+		}
+		return $p->getResult();
+    }
 }
```
imphil: please indent by four spaces. If you use Eclipse with the PDT/Symfony plugin (highly recommended), switch to the PSR-2 coding style (more or less common across PHP projects these days). I think Netbeans also has this style prepackaged.

site/src/Librecores/SiteBundle/Controller/DefaultController.php
```diff
@@ -28,6 +28,9 @@ public function homeAction(Request $request)
         // blog posts on planet
         $templateArgs['blogposts'] = $this->getBlogPosts();
 
+        $prjrepo = $this->getDoctrine()->getRepository('LibrecoresProjectRepoBundle:Project');
+		$templateArgs['activity'] = $prjrepo->findByRecentActivity(10);;
```
imphil: indentation and two `;;` at the end

site/src/Librecores/ProjectRepoBundle/Entity/ProjectRepository.php
wallento: Yes, sorry. Already did this, but after I coded that.

wallento:
Fixed review comments and merge problems.

site/src/Librecores/SiteBundle/Resources/views/Default/home.html.twig
```diff
@@ -160,63 +160,107 @@
...
+					data-livestamp="{{ project.getDateLastModified()|date("c", "UTC") }}">
+					on
+					{{ project.getDateLastModified()|date("d F Y H:i", "UTC") }}
+					UTC
+				</time>
+				.
+			</p>
+			{% endfor %}
+		</div>
+	</div>
```
imphil: Please get rid of the tabs in this file as well. HTML is usually indented with two spaces in our current code. And then rebase/squash so we can take that as a single commit. Thanks!

wallento: done

oleg-nenashev:
Please avoid moving text between lines. It messes up diffs even with the `?w=1` option enabled

wallento:
I promise I didn't change the text, but only let eclipse format the html/twig ;)

Will be more careful in future, I think its okay for that purpose now.

wallento:
I could also rebase it on top of current master to avoid that merge with resolution in it, by the way

PR #118, opened 2017-02-01
Add cookie policy

wallento:
This adds a cookie policy that must be confirmed once.

### DIFF
--- a/site/app/Resources/views/base.html.twig
+++ b/site/app/Resources/views/base.html.twig
@@ -29,27 +29,6 @@
   {% block head_additional %}{% endblock %}
 
   <link rel="search" type="application/opensearchdescription+xml" href="/search.osd" title="LibreCores" />
-
-  <link rel="stylesheet" type="text/css"
-    href="//cdnjs.cloudflare.com/ajax/libs/cookieconsent2/3.0.3/cookieconsent.min.css" />
-  <script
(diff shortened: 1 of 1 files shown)


site/app/Resources/views/base.html.twig
```diff
@@ -29,6 +29,27 @@
   {% block head_additional %}{% endblock %}
 
   <link rel="search" type="application/opensearchdescription+xml" href="/search.osd" title="LibreCores" />
+
+  <link rel="stylesheet" type="text/css"
+    href="//cdnjs.cloudflare.com/ajax/libs/cookieconsent2/3.0.3/cookieconsent.min.css" />
+  <script
+    src="//cdnjs.cloudflare.com/ajax/libs/cookieconsent2/3.0.3/cookieconsent.min.js"></script>
+  <script>
```
imphil: JS should be included at the end of the page for increased performance.

site/app/Resources/views/base.html.twig
```diff
@@ -29,6 +29,27 @@
...
+          "text": "#dddfd4"
+        },
+        "button": {
+          "background": "transparent",
+          "border": "#fae596",
+          "text": "#fae596"
+        }
+      }
+    })});
+  </script>
```
imphil: this should also be moved to the end of body part.

site/app/Resources/views/base.html.twig
wallento: Okay, sorry. I was thinking the same but then followed their instructions :)

Issue #86, opened 2016-12-24
Markdown Parser: Synchronize with GH


imphil:
Our MarkDown parser seems to be slightly different than the GitHub one.

See this project:
https://github.com/kactus2/kactus2dev/blob/master/README.md
https://www.librecores.org/epekkar/kactus2

imphil:
I guess we'll never be able to get bug-free output unless we go with the exact same solution as GH uses (https://github.com/github/markup). 

Idea:
- Switch the Project::descriptionText and Project::licenseText fields to contain processed HTML.
- wrap that Ruby stuff in a separate script that we call when processing.
- Switch from editing MarkDown to HTML on LibreCores itself.

olofk:
As you say, we will never have bug-free output unless using the same solution as github. The problem is that this isn't enough, as the text might contain internal links that will be broken. The source could also be asciidoc or Restructured Text. (For the latter, I have a similar problem with the README.rst in FuseSoC that is also used by pypi, and has broken links there). Wouldn't the easiest solution be to do a dumb import, but give users the chance to edit and fix-up things. I think it will be a lot of effort spent in vain, trying to emulate other parsers, and editing the raw md/adoc/rst is IMO much nicer than preprocessed html

berndca:
The a HTML version rendered by GitHub's Markup can be obtained through the AGitHub API see #116 for an example.

@olofk Just change all your links to full URL's and they will work everywhere. For instance replace
```doc/fusesoc.adoc``` with ```https://github.com/olofk/fusesoc/blob/master/doc/fusesoc.adoc```

imphil:
done in PR #123

imphil:
For example https://www.librecores.org/epekkar/kactus2 shows up correctly now.

Issue #85, opened 2016-12-23
Better handle huge repositories in UpdateProjectInformation


imphil:
Example: https://github.com/krevanth/ZAP is more than 300 MB and cannot be cloned in 3 minutes. We probably should do a couple of things for such repos: 

- do a flat clone to get the README and LICENSE file.
- re-use a clone made once with much higher timeout (also look into Process::setIdleTimeout()) and update that periodically to get the latest repo statistics.
- other clone options to get only essential parts and not the full repo data which we in the end don't care about?

oleg-nenashev:
Maybe there is no need to clone repositories at all (in the case of projects on GitHub). It is possible to use GitHub API to get file listings and then to download files directly

berndca:
Please note that the archive for this repo https://github.com/krevanth/ZAP/archive/master.zip is less than 1MB and it downloads in a few seconds. Of course this will only work when the "download as archive" option is available.

imphil:
We're now using a flat clone, this solved the issue for now. Everything else will involve the API integration, which is coming in the future (and different issues).

PR #119, opened 2017-02-01
Remove repo checkout in case of exception

wallento:


imphil:
I'm fully refactoring this code to support #86 and #85 and similar issues, so this code is going away anyways in the next couple of days. Thanks anyways!

wallento:
Ah, okay, I have some input regarding that that was probably already mentioned somewhere: After playing with the code I come to the conclusion that SourceRepo is maybe the best abstraction layer for Git, GitHub etc. So it could provide the proxy to a README, statistics, etc. Maybe we should have a discussion in our call about that.

imphil:
That abstraction is actually the reason I'm redesigning this code. It will enable us to support different APIs (like GitHub), or fall back to our own parser. More information to come when I get to finish this stuff the next couple of days.

Issue #65, opened 2016-10-09
Add support for user and organization pictures/icons


imphil:
Every user/org should be associated with a picture.

Give users the option:
- pick existing one from gravatar (default)
- upload an individual image

eliaskousk:
For Gravatar the [Gravatar Bundle](https://github.com/henrikbjorn/GravatarBundle) seems good.

For manual uploading we can use the [Vich Uploader Bundle](https://github.com/dustin10/VichUploaderBundle).

It has a somewhat complex configuration that requires a separate fiesystem / storage abstraction e.g [Gaufrette](https://github.com/KnpLabs/KnpGaufretteBundle) (it supports AWS) but I've used it in the past and it works fine. It will also allow us to do other kinds of uploads if and when needed.

Let me know if you have any better suggestions since it was more than 6 months ago that I scouted for these bundles. Maybe something newer / better / more maintained / more secure has come along since then.

oleg-nenashev:
> upload an individual image

Why would we actually need it? I'm pretty sure Gravatar is enough for MVP, and actually dialing with image requirements is something taking much time without so much outcome.

imphil:
@oleg-nenashev mostly for organizations, which does not have an email address assigned.

PR #123, opened 2017-02-04
Rework repo parsing

imphil:
@eliaskousk can you have a look at the general design here quickly? For now it just takes the existing code and refactors it, but enables us to use e.g. the GitHub API instead of parsing the repo ourselves in the future.

The major visible difference is now that the license and readme are HTML and converted from the markup content using the exact same tooling as github uses, making the output much closer to what github presents. [Side note: In the future we can avoid *for GitHub* converting ourselves by using the GitHub API; but then we still need this code for other repo providers]


berndca:
> [Side note: In the future we can avoid for GitHub converting ourselves by using the GitHub API; but then we still need this code for other repo providers]

You can use the GitHub API to convert **ANY** arbitrary text to HTML using their Markup renderer.

imphil:
@berndca do you mean this API? https://developer.github.com/v3/markdown/ This works only for Markdown. 

For README and other content inside a GitHub repository, there is https://developer.github.com/v3/repos/contents/#get-the-readme which also renders other markup formats to HTML. 

However, let's keep this PR focused on the issue itself, this question is tangential to this. (The code structure allows to make such a change in the future easily.)

berndca:
I tried both the fusesoc REDME.rst and the README.md from kactus2. I copied the text from the raw view into the body of the post request and got the expected HTML. I agree that they could be a little more specific in their documentation. It does say however: 

> The rendering mode. Can be either:
> * markdown to render a document as plain Markdown, just like README files are rendered.

berndca:
I am concerned about replacing the markdown editor with trumbowyg. IMO this will make the site less secure. You can manipulate the HTML output of trumbowyg easily and use it to create XSS vulnerability. Most attacks use broken HTML which is notoriously hard to pass. Browsers tolerate broken HTML however their response varies not only by manufacturer but also by release. Accepting HTML from the user is a dangerous proposition. The safest way is to use something like markdown and just brutally escape the whole text, i.e. not allow any raw HTML no matter how trivial.

There are other ways to style user content.

berndca:
I could not find any code obtaining GitHub meta data like number of open issues etc. Forgive my ignorance, did I just not look at the right places?

imphil:
Two points:

- First, we need a solution which is able to edit any content we get from the README in the repository. The previous solution converted all content (such as POD and Asciidoc) to Markdown. This process is lossy, as Markup is less expressive than HTML. So the only common superset is HTML, which is what I went for. 

- Second, we use HTML Purifier on all HTML content before it is stored or displayed anywhere. This whitelist-based solution has a very good track record, is used widely and generally agreed on to be safe. (Of course, just like with any software, the general possibility of security-relevant bugs remains.)

imphil:
> I could not find any code obtaining GitHub meta data like number of open issues etc. Forgive my ignorance, did I just not look at the right places?

no,there isn't any yet. What made you think it should be there?

berndca:
> no,there isn't any yet. What made you think it should be there?

The title of this PR and the impression that there was an agreement om #65 and #80.

imphil:
The PR title is "Rework repo parsing" -- no GitHub there. 

And from the description:
> For now it just takes the existing code and refactors it, but enables us to use e.g. the GitHub API instead of parsing the repo ourselves in the future.

We'll always need a fallback path to our own repo parsing. Not everybody is using GitHub. This was a very early design decision which still holds.

berndca:
I'm puzzled why you would want to edit the README. Would you not prefer that any updates, changes etc. made to the README are done in it's repository? If there is any information (annotations etc.) which should be added I would think that it would be prudent to show this information separate from the original repo.

site/app/AppKernel.php
```diff
@@ -16,7 +16,6 @@ public function registerBundles()
             new Symfony\Bundle\AsseticBundle\AsseticBundle(),
             new Doctrine\Bundle\DoctrineBundle\DoctrineBundle(),
             new Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle(),
-            new Knp\Bundle\MarkdownBundle\KnpMarkdownBundle(),
```
eliaskousk: We can't remove the markdown bundle since our static page content wrapper template uses the markdown filter.

ansible/roles/db/tasks/librecores-blog.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Create librecores-blog MySQL database
-  mysql_db: name=librecores-blog state=present collation=utf8_general_ci encoding=utf8
+  mysql_db: name=librecores-blog state=present collation=utf8mb4_unicode_520_ci encoding=utf8mb4
```
eliaskousk: Good idea, Greek is also a strange language!

ansible/roles/web/tasks/github-markup.yml
```diff
@@ -0,0 +1,34 @@
+---
+# github-markup tool
```
eliaskousk: Although I don't like having any ruby dependencies, I also think that this is for the best and will allow us to be consistent with the way Github renders the files.

site/composer.json
```diff
@@ -24,16 +24,14 @@
 		"sensio/distribution-bundle" : "~3.0,>=3.0.12",
 		"sensio/framework-extra-bundle" : "~3.0,>=3.0.2",
 		"incenteev/composer-parameter-handler" : "~2.0",
-		"knplabs/knp-markdown-bundle" : "~1.3",
```
eliaskousk: See comment on AppKernel.

site/src/Librecores/ProjectRepoBundle/Entity/Project.php
```diff
@@ -126,7 +129,8 @@ class Project
      *
      * @var string
      *
-     * @ORM\Column(type="text", nullable=true)
+     * @Assert\Length(max = 50000)
```
eliaskousk: I agree with the size. (it's almost perfect - 64KB is the perfect one for obvious reasons!)

eliaskousk:
@berndca I understand your point that it's best for the project maintainer to change the readme on their original repository instead of changing it on our web-app. However there is a legitimate use case for such edits (and I might be missing others since I wasn't around when this design decision was taken) :

What if somehow nobody has access to the original repo to change the readme and there is something important that needs to be said on the readme that is shown on librecores regarding this project? I know that some hardware projects can become pretty unmaintained and there needs to be a way to present important information for such projects on librecores.

Yes there are other solutions like forking the repo somewhere and properly changing files there. But I also think it's best if one, possibly someone who can't bother to do the transfer and just wants to present an old core to users of librecores, can resort to this quick and easy but "bad" practice of editing files outside the repo and inside librecores. I think the new system is pretty flexible so everybody's needs and time constraints when maintaining an old ip core / hdl library should be covered, whether or not they have access to the original repo.

site/app/AppKernel.php
imphil: Good catch, thanks.

site/src/Librecores/ProjectRepoBundle/Entity/Project.php
imphil: 64k it be :)

berndca:
@eliaskousk Thanks for taking the time to walk me through the history.

Issue #133, opened 2017-03-05
Using editorconfig for consistent styles


Agathver:
[Editorconfig](http://editorconfig.org/) is a cross IDE code style configuration. Since everyone has their favorite IDE, editorconfig will automatically set appropriate indentation and line ending standards consistently across all supported IDEs.

PR #134, opened 2017-03-15
Add editorconfig

Agathver:
An editorconfig will help to synchronize  whitespace settings such as tab size, indentation type across all editors. This will help new contributors in writing code that better adheres to the code style of the rest of the repo. ( #133 )

This pull request creates an editorconfig file for the project with the following settings. It closes issue #133 

| Setting | Value |
| --------- | ------- |
| Indent style | `space` |
| Indent size | 2 |
| Line endings | `lf` |
| Character encoding | `utf-8` |
| Trim trailing whitespace | yes |
| Insert final newline | yes |

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[{package.json,*.yml}]
+indent_style = space
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,9 +8,5 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[{package.json,*.yml}]
-indent_style = space
-indent_size = 2
-
 [*.md]
 trim_trailing_whitespace = false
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[{package.json,*.yml}]
+indent_style = space
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = false
```
imphil: this should be `true` as well

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[{package.json,*.yml}]
+indent_style = space
+indent_size = 2
```
imphil: isn't that the same as the default?

.editorconfig
Agathver: Two trailing spaces signify a `<br/>`, setting this to true will break markdown in many places.

imphil: ah ok, I see.

imphil:
Now only the PHP part is missing and we're done.

Agathver:
> Now only the PHP part is missing and we're done.

Our PHP files use 4 spaces right ?

imphil:
Hm, it seems github dropped part of my review from the beginning. For PHP, we follow PSR-2, so something like that should do it: https://github.com/pfefferle/.editorconfigs/blob/master/php-psr-2/.editorconfig

Given that everything except for the indent size is already default, your change is all that's needed. Thanks!

PR #138, opened 2017-05-04
Add OSDDI to planet

olofk:


planet/config.ini
```diff
@@ -65,6 +65,9 @@ name = Kestrel Computer Project
 [http://blog.verificationgentleman.com/feeds/posts/default]
 name = Verification Gentleman
 
+[http://abopen.com/blog/category/osddi/feed]
```
imphil: We're trying to avoid adding more HTTP blogs to the feed, as they cause mixed-content warnings when they contain images. 

It seems the ssl certificate for abopen.com has expired on 05.01.2016, so some action is required.

Andrew, is HTTPS support on your todo list for the short term? Otherwise I'm going to add the blog with HTTP for now and then we can change it later on.

9600: Ah, that's been on my to-do list for some time. Yes, if you could use HTTP for now and I'll try and fix at some point over the next few weeks. Lots of vhosts on that system and I need to revisit the config.

PR #141, opened 2017-05-25
Add blog feed to planet

Agathver:
Link to an atom feed on my blog for all posts about LibreCores.org development during GSoC 2017


imphil:
Thanks, I'll push it to production soon!

PR #147, opened 2017-06-15
Add phpunit testing and travis CI

### LINKED ISSUE #146: Testing and CI
We need to implement unit testing and integration testing for the Symfony based librecores.org site.

1. Using PHPUnit for units tests.
2. Symfony PHPUnit Bridge and Symfony BrowserKit for integration testing.
3. Travis CI for continous integration

Agathver:
Adds symfony php-unit bridge and phpunit, also configures Travis CI

Current CI report: https://travis-ci.org/Agathver/librecores-web

Closes #146

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 LibreCores Web Site
 ===================
 
+**Website**: [![Build Status](https://travis-ci.org/librecores/librecores-web.svg?branch=master)](https://travis-ci.org/librecores/librecores-web)
+
 This is the code running on https://www.librecores.org.
 
 The LibreCores web site is written in PHP and uses the Symfony framework.
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,23 @@
+dist: trusty
+sudo: false
+
+language: php
+php:
+  - '7.1'
+env:
+  - SYMFONY_ENV=test
+
+git:
+  depth: 1
+
+cache:
+  directories:
+   - site/vendor
+
+install:
+  - cd site && composer install --no-scripts --ignore-platform-reqs
+
+script:
+  - ./vendor/bin/phpunit
+
+
--- a/site/app/config/config_test.yml
+++ b/site/app/config/config_test.yml
@@ -1,6 +1,13 @@
 imports:
     - { resource: config_dev.yml }
 
+parameters:
+    # Dummy OAuth API keys for testing
+    github_app_id:     DUMMY
+    github_app_secret: DUMMY
+    google_app_id:     DUMMY
+    google_app_secret: DUMMY
+
 framework:
     test: ~
     session:
--- a/site/composer.json
+++ b/site/composer.json
@@ -70,5 +70,9 @@
 		"branch-alias" : {
 			"dev-master" : "2.8-dev"
 		}
+	},
+	"require-dev": {
+		"phpunit/phpunit": "^6.2",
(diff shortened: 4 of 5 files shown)


oleg-nenashev:
Is it really required to keep `composer.lock` in the repo? Is it possible to regenerate it on-demand from composer.json?

eliaskousk:
@oleg-nenashev I personally don't like keeping the .lock file in the repo and this is usually recommended only for libraries (the fact that I don't write libraries is of-topic :P).

However, for applications such us ours, the general consensus in the php / composer community is that it's better to keep it in the repo and thus lock all dependency versions explicitly in-case we haven't checked / selected the semver of each library thoroughly. And of-course just in case one of the libraries our app uses doesn't follow sem.ver 100%.

Oh and in short the way each one (.json / .lock) is used by composer is the following:
- composer install uses the versions from the .lock file and if it can't find it fallbacks to the .json one
- composer update uses the versions from the .lock file always and then updates the .lock file with anything newer

README.md
```diff
@@ -1,6 +1,8 @@
 LibreCores Web Site
 ===================
 
+**Website**: [![Build Status](https://travis-ci.org/Agathver/librecores-web.svg?branch=master)](https://travis-ci.org/Agathver/librecores-web)
```
imphil: that's a old link?

site/app/config/parameters.yml.dist
```diff
@@ -27,11 +27,11 @@ parameters:
...
+    github_app_id:     GITHUB_OAUTH_ID
+    github_app_secret: GITHUB_OAUTH_SECRET
 
     # OAuth configuration: Google
     # Register at https://console.developers.google.com/
     # https://console.developers.google.com/project/librecores/apiui/credential/oauthclient
-    google_app_id:     ~
-    google_app_secret: ~
+    google_app_id:     GOOGLE_OAUTH_ID
+    google_app_secret: GOOGLE_OAUTH_SECRET
```
imphil: What is this change for?

imphil:
I've now enabled Travis integration for this repo, you can push again with your changes to see a CI run also in this PR.

site/app/config/parameters.yml.dist
Agathver: We generate `parameters.yml` from `parameters.yml.dist` directly in CI. `HWIOAuthBundle` requires the api keys to be not null.

imphil: where? I don't see it in the `.travis.yml`. Either way, the correct solution is not to copy the .dist file, but to create a new file based on this template.

Agathver: In the absence of an interactive terminal, symfony parameter generator will directly copy the `.dist` file. I am looking into an alternate way, involving Symfony environments

imphil: ah I see. Then probably the easiest path forward is to add a comment before the dummy values to hint people to replace it, and leave the change as-is otherwise.

Agathver:
@oleg-nenashev IMO, we should commit the `composer.lock`. It is essential for every packages to be deterministic - dev, CI and production should run the same version of all dependencies. Composer guarantees this using the lock file.

Without the `composer.lock`, it will also be impossible to perform a `composer install` inside Travis, since it requires a lot of RAM to resolve dependencies.

imphil:
Nice work, thanks!

PR #150, opened 2017-06-16
Extract and display project activity metrics

agathver:
Displays and extracts metrics after git clone.

- [x] Extract commits
- [x] Extract contributors
- [x] Display commit count
- [x] Display contributors
- [x] Histograms for commits
- [x] Extract lines of code and comments
- [x] Display LOC
- [x] Tests
- [x] Code-style fixes


site/src/Librecores/ProjectRepoBundle/Entity/Project.php
```diff
@@ -47,7 +47,7 @@ class Project
      * @var string
      *
      * @Assert\Choice(choices = {"ASSIGNED", "UNASSIGNED"})
-     * @ORM\Column(type="string")
+     * @ORM\Column(type="string", options={"default" : self::STATUS_ASSIGNED})
```
imphil: That's causing  the following error during provisioning:

```
An error occurred when executing the \"'cache:clear --no-warmup'\" command:                                                                                                                                                                                                                                               ", "  PHP Fatal error:  Uncaught Doctrine\\Common\\Annotations\\AnnotationException: [Semantical Error] Couldn't find constant self::STATUS_ASSIGNED, property Librecores\\ProjectRepoBundle\\Entity\\Project::$status. in /var/www/lc/site/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationException.php:54  ", "  Stack trace:                                                                                                                                                                                                                                                                                                            ", "  #0 /var/www/lc/site/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationException.php(101): Doctrine\\Common\\Annotations\\AnnotationException::semanticalError('Couldn't find c...')                                                                                                                    ", "  #1 /var/www/lc/site/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php(973): Doctrine\\Common\\Annotations\\AnnotationException::semanticalErrorConstants('self::STATUS_AS...', 'property Librec...')                                                                                               ", "  #2 /var/www/lc/site/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php(1039): Doctrine\\Common\\Annotations\\DocParser->Constant()                                                                                                                                                                  ", "  #3 /var/www/lc/site/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php(1161): Doctrine\\Common\\Annotations\\DocParser->PlainValue() in /var/www/lc/site/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationException.php on line 54
```

imphil: Switching to "ASSIGNED" instead of using a constant works.

imphil:
It also seems this PR is missing a database migration to update the DB to the latest schema, including the added tables. Please add that and run
```
vagrant destroy
vagrant up
```

to test if the provisioning, including loading of the db fixtures, works as expected.

imphil:
The deployment in Vagrant works now, thanks. 
But I still don't get a fully working version: it seems the chart js library is not properly loaded. The developer console says: 

```
01:31:52.740 TypeError: $chart.height(...).peity is not a function 1 optimsoc:227:5
	<anonym> http://www.librecores.devel/test2/optimsoc:227:5
	m.Callbacks/j http://www.librecores.devel/js/app_body_part_1.js:2:27304
	m.Callbacks/k.fireWith http://www.librecores.devel/js/app_body_part_1.js:2:28122
	.ready http://www.librecores.devel/js/app_body_part_1.js:2:29954
	J http://www.librecores.devel/js/app_body_part_1.js:2:30320
```

This is how it looks in Vagrant on my machine:
grafik

Also, the JS code using `$chart` as variable looks very much inspired by PHP code. JS variables don't need/use `$` in the beginning.

Some other comments:
- Why does it say it would take 12 months by 97 developers -- where are the 97 developers coming from? (since there have been 13 contributors, as shown below.)
- Please don't directly render relative times, but use the absolute time (in UTC) and let JS make a relative time out of it. See https://github.com/librecores/librecores-web/blob/master/site/src/Librecores/PlanetBundle/Resources/views/Default/index.html.twig#L20 (and don't forget the tooltip code at the end of the page.)
- The optimsoc example also clearly shows why it would be useful to group contributions not only by email address, but also by name. We left this explicitly out of scope, but please create another issue as enhancement to track this for the future.
- The UI in the datasheet is pretty crowded at the moment, but we're already working on that in the next iterations. So that's fine for now.


I'll have a closer look at the code once the functionality works for me to make sure we don't waste time reviewing code which will change anyways.

Agathver:
@imphil 
>it seems the chart js library is not properly loaded.

I cannot reproduce it. It is the part of app_body_1.js, I can verify that, and the file isn't missing from the commits. Looks like a browser cache issue to me. (I recall seeing `expire: max` in nginx config). 

Anyways, I'll add a cache-busting filter to Assetic, we will surely need it in production

imphil:
> I cannot reproduce it. It is the part of app_body_1.js, I can verify that, and the file isn't missing from the commits. Looks like a browser cache issue to me. (I recall seeing expire: max in nginx config).
>
> I'll add a cache-busting filter to Assetic, we will surely need it in production es

There is indeed a bug somewere on my setup, but its most likely not the caching of nginx. Instead, it seems the assets don't get regenerated properly:

```
~/src/librecores-web/site/web/js> ll
total 728
-rw-r--r-- 1 philipp users   7228 Apr 28  2016 8864e78.js
-rw-rw-r-- 1 philipp users 239184 Sep 28  2016 app_body.js
-rw-rw-r-- 1 philipp users 239184 Sep 28  2016 app_body_part_1.js
-rw-rw-r-- 1 philipp users   7228 Sep 28  2016 app_head.js
-rw-rw-r-- 1 philipp users   7228 Sep 28  2016 app_head_part_1.js
-rw-r--r-- 1 philipp users 236240 Apr 28  2016 f0e7c52.js
```

I'll look into it.

site/src/Librecores/ProjectRepoBundle/Util/DefaultExecutor.php
```diff
@@ -0,0 +1,83 @@
...
+use Symfony\Component\Process\ProcessBuilder;
+
+/**
+ * Default implementation of ExecutorInterface
+ *
+ * Uses Symfony ProcessBuilder component to build the process.
+ *
+ * @author Amitosh Swain Mahapatra <amitosh.swain@gmail.com>
+ */
+class DefaultExecutor implements ExecutorInterface
```
imphil: What other ExecutorInterface instances do you expect in the future? Right now, this class hides code and hence makes it less explicit and readable. For example, 
`$builder->setWorkingDirectory(dir)` is much clearer than `exec(..., ['cwd' => dir], ...)`. At the same time, the IDE is able to help finding the right method.

site/src/Librecores/ProjectRepoBundle/Resources/views/Project/view.html.twig
```diff
@@ -104,3 +158,11 @@
 
 </div>
 {% endblock %}
+
+{% block pagejs %}
+  $(function(){
+    $('[data-toggle="tooltip"]').tooltip();
+    var $chart = $('.librecores-project-activity-graph');
+    $chart.height('100px').peity('line', { height: $chart.height(), width: $chart.width() });
```
imphil: Please don't use `$chart` as variable name in JS, it's not PHP.

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -164,4 +179,124 @@ public function getLicenseTextSafeHtml(): ?string
...
+
+        $this->logger->debug("Fetched commits from $cwd");
+        $this->parseCommits($output);
+    }
+
+    /**
+     * {@inheritDoc}
+     * @see RepoCrawler::commitExists()
+     */
+    public function commitExists(string $id) : bool
```
imphil: $id => $commitId?

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -164,4 +179,124 @@ public function getLicenseTextSafeHtml(): ?string
...
+        {
+            case 0:     // successful, commit exists
+            case 1:     // commit does not exist in branch
+            case 128:   // commit does not exist in repository
+                $this->logger->debug("Checked commits from $cwd");
+                return $code === 0;
+            default:    // anything other than 0 or 1 is error
+                //die($code);
+                throw new \RuntimeException("Unable to fetch commits from $cwd : ".$errors);
+        }
```
imphil: This switch statement needs refactoring. "// anything other than 0 or 1 is error" doesn't match reality, and "return $code === 0;" together with the fall-through case statements is hard to understand. Spend a line more code to make it readable.

site/src/Librecores/ProjectRepoBundle/RepoCrawler/RepoCrawler.php
```diff
@@ -13,16 +17,60 @@
...
-        MarkupToHtmlConverter $markupConverter, LoggerInterface $logger)
+    /**
+     * @var ExecutorInterface
+     */
+    protected $executor;
+
+    /**
+     * @var SourceCrawlerInterface[]
+     */
+    protected $sourceCrawlers;
```
imphil: Now thats rather confusing. We have a RepoCrawler which itself has SourceCrawlers? How are the tasks between the two separated, and how can this be reflected in the class/interface names?

This class is now (and partially has been) a mix between an actor and an entity representing an repository. Let's continue the discussion on the mailing list where you already came up with a first solution.

site/src/Librecores/ProjectRepoBundle/Util/Dates.php
```diff
@@ -0,0 +1,70 @@
...
+namespace Librecores\ProjectRepoBundle\Util;
+
+use Symfony\Component\Finder\Iterator\DateRangeFilterIterator;
+
+/**
+ * Utility methods for dates
+ *
+ * @author Amitosh Swain Mahapatra <amitosh.swain@gmail.com>
+ */
+class Dates
```
imphil: Can we find a more descriptive class name here?

site/src/Librecores/ProjectRepoBundle/Util/DefaultExecutor.php
Agathver: Lambda based executors ? This also makes testing easier. Using ProcessBuilder made it impossible to test just the algorithm by faking output.

imphil: First: You mean AWS Lambda? It's an unlikely assumption we will only execute a command on lambda and all other processing will be done on another machine. The more likely point of decoupling is the RabbitMQ queue, i.e. the full crawler. In this case, no decoupling of the command execution is needed.

Second on testing: usually, the recommended textbook advice is to create two classes, one to obtain the data (in this case by calling a function, in other cases it would be from an API) and the other one to use this data in some way. The first class is not unit tested, but covered by integration tests and for the purposes of unit tests assumed to be working. The second class (which does the processing and contains the relevant custom code) is the one tested by unit tests.

Now, we have a bit of a mixture between getting data and processing it in the crawler classes. That's fine, as it leads to efficient and readable code. Testing this setup is also easy: create a shell script called "git", which you inject into the `PATH` environment variable before running tests. Then you can let the shell script return your golden output, just like git would, and be done with it. No large mocking framework or overloaded class hierarchy needed.

site/src/Librecores/ProjectRepoBundle/Doctrine/DefaultMetadataManager.php
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * DefaultMetadataManager constructor.
+     * @param CommitRepository $commitRepository
+     * @param ContributorRepository $contributorRepository
+     */
+    public function __construct(
+        CommitRepository $commitRepository,
+        ContributorRepository $contributorRepository
+    ) {
```
agathver: This is PSR-2 . Should I change it ?

agathver:
@oleg-nenashev I have disabled COCOMO metrics display

agathver:
@imphil @oleg-nenashev I have fixed the issues, please review again

site/src/Librecores/ProjectRepoBundle/Doctrine/Hydrators/GroupHydrator.php
```diff
@@ -0,0 +1,53 @@
...
+            $result[$row[0]][] = count($value) > 1 ? $value : $value[0];
+        }
+
+        foreach ($result as $key => $item) {
+            $result[$key] = $this->group($item);
+        }
+
+        return $result;
+    }
+}
```
agathver: Resolved

imphil:
Merged, thanks @agathver. I've seen a couple minor things, but let's do that in subsequent iterations.

oleg-nenashev:
> I've seen a couple minor things, but let's do that in subsequent iterations.

Fine for me

PR #151, opened 2017-06-20
Add Docker packaging for the development environment

oleg-nenashev:
Work in progress...

Currently it fails at the `TASK [web : reload systemd to pick up changes in Unit files] stage` due to the missing XDG stuff for the `root` user. Will likely need to use another user.

- [ ] Make it working
- [ ] Document it
- [ ] Add a Docker-based test of the Ansible logic to the PR builder

### DIFF
--- a/ansible/roles/web/tasks/librecores-site.yml
+++ b/ansible/roles/web/tasks/librecores-site.yml
@@ -53,17 +53,20 @@
 # we need to work around ansible trying to chown a file on NFS, which
 # is not possible. See https://github.com/ansible/ansible/issues/7372
 
+- name: Create directory for generated Symfony files
+  command: mkdir -p /opt/lc/site/app/config/
+
 - name: Create Symfony environment shell file (part I, populate template in temp folder)
   template: src=symfony-env.sh.j2 dest=/tmp/symfony-env.sh
 
 - name: Create Symfony environment shell file (part II, move to destination)
-  command: mv /tmp/symfony-env.sh /var/www/lc/site/app/config/symfony-env.sh
+  command: mv /tmp/symfony-env.sh /opt/lc/site/app/config/symfony-env.sh
 
 - name: Create Symfony configuration (part I, populate template in temp folder)
   template: src=parameters.yml.j2 dest=/tmp/parameters.yml
 
 - name: Create symfony configuration (part II, move to destination)
-  command: mv /tmp/parameters.yml /var/www/lc/site/app/config/parameters.yml
+  command: mv /tmp/parameters.yml /opt/lc/site/app/config/parameters.yml
 
 - name: "Create composer configuration directory for user {{ web_user }}"
   file:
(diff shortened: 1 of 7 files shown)


ansible/roles/web/tasks/librecores-site.yml
```diff
@@ -53,17 +53,20 @@
...
 # is not possible. See https://github.com/ansible/ansible/issues/7372
 
+- name: Create directory for generated Symfony files
+  command: mkdir -p /opt/lc/site/app/config/
+
 - name: Create Symfony environment shell file (part I, populate template in temp folder)
   template: src=symfony-env.sh.j2 dest=/tmp/symfony-env.sh
 
 - name: Create Symfony environment shell file (part II, move to destination)
-  command: mv /tmp/symfony-env.sh /var/www/lc/site/app/config/symfony-env.sh
```
oleg-nenashev: It will also impact Vagrant and production packaging

docker/run.sh
```diff
@@ -0,0 +1,20 @@
...
+set -x
+cd /var/www/lc/site
+
+# Install dependencies through composer
+composer install
+
+# Migrate database
+php bin/console doctrine:migrations:migrate -n
+#sudo_user: "{{ web_user }}"
+#environment: "{{ symfony_config }}"
```
oleg-nenashev: TODO: this environment needs to be sourced here

Agathver:
I made a similar effort, but with a different approach - with every service running in its own container. 
Docker community frowns on running multiple services in a single container. This kind of setup is easy to manage and test (ref: Wercker)

The major issue I faced is file ownership gets messed up while using containers, while performing a `bind mount` to share the source . NFS is out of question and FUSE mounts like sshfs are also not possible due to lack of FUSE access inside containers.

LXC can also be an easier alternative, (above demerits apply): https://gist.github.com/Agathver/71b8759805632c432fef1665e931e303

oleg-nenashev:
Having a single image is useful because then you can do the stuff without Vagrant tools on the local machine. The intention is also to reuse the existing Ansible startup logic in order to avoid duplications. Otherwise I would just use Docker compose.

> The major issue I faced is file ownership gets messed up while using containers, while performing a bind mount to share the source . NFS is out of question and FUSE mounts like sshfs are also not possible due to lack of FUSE access inside containers.

In Docker [Compose] world it's most simple to share files via a shared volume. NFS cannot be launched easily in Docker since it requires some extra core components, but CIFS can be an option.

Agathver:
Added #153 to track this one

oleg-nenashev:
I have not pushed the working version yet. It's working, but it causes some regressions in the Vagrant flow

imphil:
One challenge when getting our ansible deployments working in a container is the various services that make up librecores-web. The easiest solution would be to continue using systemd as PID1 and leave the deployment scripts as they are for Vagrant and AWS. Using a base image like https://hub.docker.com/r/solita/ubuntu-systemd/ should make that work. Since we're doing that only for testing, we don't need to put much effort into splitting each service into a different container using docker-compose or similar tools like ansible-container.

oleg-nenashev:
The referenced container is a bad idea, because it requires the privileged mode. I still hope to workaround it without creating much mess around

imphil:
I'm wondering how RHEL solves this? Judging from https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/managing_containers/using_systemd_with_containers they are able to run systemd without privileged containers? However, judging from the description, the service files are modified?

oleg-nenashev:
Dev version is here: https://hub.docker.com/r/librecores/librecores-web-dev/

Issue #148, opened 2017-06-15
Tests for services and routes


Agathver:
We have a large number of routes and services and no tests for them. We should start adding tests.

Agathver:
This is an extensive task, so we should break it into phases:

1. Functional tests to make sure, at least all routes are displayed correctly
2. Unit tests for services
3. Functional tests for all routes and forms

PR #152, opened 2017-06-21
Add functional tests for routes

Agathver:
This PR ~~sets up Travis CI for~~ adds functional tests ~~involving MySQL and RabbitMQ~~. 

It adds a test for the DefaultController, which should be used as a template for writing further tests.

#148

### DIFF
--- a/site/tests/Librecores/PlanetRepoBundle/Controller/DefaultControllerTest.php
+++ b/site/tests/Librecores/PlanetRepoBundle/Controller/DefaultControllerTest.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Tests\Librecores\ProjectRepoBundle\Controller;
+
+
+use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\ORM\EntityRepository;
+use FOS\UserBundle\Doctrine\UserManager;
+use Librecores\ProjectRepoBundle\Entity\User;
+use Symfony\Bridge\Doctrine\RegistryInterface;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+
+/**
+ * Test for DefaultController
+ *
+ * @author Amitosh Swain Mahapatra <amitosh.swain@gmail.com>
+ */
+class DefaultControllerTest extends WebTestCase
+{
+    public function testUserOrOrgViewAction()
+    {
+        // We use separate environments for testing in local and CI
+        // if env var SYMFONY_ENV exists, then pass it as options to kernel builder
+        // else fallback to default. This is required for all tests
+
+        $env = getenv('SYMFONY_ENV');
+        $client = static::createClient($env ? [ 'environment' => $env ] : []);
+
+        // test for a user
+
+        $crawler = $client->request('GET', '/test');
+
+        $this->assertEquals($client->getResponse()->getStatusCode(), 200);
+
+        // ensure header contains user's name
+        $this->assertEquals($crawler->filter('.librecores-user-page-header > h1')->eq(0)->text(),'test');
+
+        $crawler = $client->request('GET', '/openrisc');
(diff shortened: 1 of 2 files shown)


imphil:
This commit recreates part of the deployment rules we have in Ansible today inside the travis configuration. I'm not convinced this is a viable path in the longer term: 
- Deployment is tricky: getting all required components of the deployment right isn't easy.
- We want to test what we actually use and not fight bugs caused by discrepancies between different testing environments.

In my opinion a more viable long-term approach would be:
- Add tests and testing frameworks as needed and ensure that testing locally in the standard Vagrant environment works.
For test automation, we then have two options:
- To test in travis, wait for the docker deployments to be ready. #151 
- If that takes too long or doesn't work out, we can always use a baremetal machine with LibreCores CI (Jenkins) to run the tests in our own infrastructure using the already existing Vagrant setup.

Agathver:
@imphil The main aim of these tests is to make sure that a PR accidentally does not break any feature due to modifying components which other parts are dependent on. This does not aim to be a full replacement for an "integration test" for which we will continue to need a staging box.

What it aims to do:
1. Verifies that there is no errors in controllers and that they return a correct response in a mocked/controlled environment,
2. Services work as expected.
3. Database queries work as expected.

For testing all the parts correctly, we undoubtedly need Selenium or equivalent on a staging environment configured identical to the production environment.

imphil:
I agree with the goals, just the deployment part inside travis.yml replicating our setup is pretty fragile in my opinion. So let's do it this way:
- First, split up this commit into a part adding the tests and a part adding the travis deployment.
- The first part can go in immediately, no concerns here.
- For the second part, the travis integration, I'm proposing two things:
  - We can add it as temporary stopgap measure until we have docker in travis running. Looking at the PR #151 it seems that's not in the far future. Also, don't put any more effort into it making the environment nicer.
  - Open a follow-up issue and to remove this stopgap again and replace it as soon as we either have docker in travis running, or we have another solution ready.

I hope this approach doesn't block your development process, but also makes it clear that this solution won't stay (and shouldn't stay).

site/tests/Librecores/SiteBundle/Controller/tests/Librecores/SiteBundle/Controller/DefaultControllerTest.php
```diff
@@ -0,0 +1,40 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: amitosh
+ * Date: 21/6/17
+ * Time: 3:05 PM
+ */
+
```
imphil: You can remove this header.

site/tests/Librecores/SiteBundle/Controller/tests/Librecores/SiteBundle/Controller/DefaultControllerTest.php
```diff
@@ -0,0 +1,40 @@
...
+        $env = getenv('SYMFONY_ENV');
+        $client = static::createClient($env ? [ 'environment' => $env ] : []);
+
+        $crawler = $client->request('GET', '/');
+
+        $this->assertEquals($client->getResponse()->getStatusCode(), 200);  // must be a 200 OK response
+
+        // since most of the content is static, extensive testing is not required, asserting presence of certain key elements
+        $this->assertEquals($crawler->filter('.librecores-home-banner-wordmark')->html(), 'LibreCores');
+        $this->assertEquals($crawler->filter('.librecores-home-banner-claim')->html(), 'We drive Free and Open Digital Hardware.');   //expect certain values
```
imphil: Please keep the lines around 80 chars (not a strict requirement, but roughly).

Agathver:
@imphil I have removed changes to Travis CI that dealt with setting up MySQL and RabbitMQ

imphil:
Merging this for now as it's mostly example code anyways. But please keep the lines around 80 characters. This makes reviewing and editing much easier.

Issue #154, opened 2017-07-05
Extract repository metrics from git hosting provider APIs


agathver:
Github and Gitlab provide extensive APIs to extract repository metadata, in addition to metrics such as issues and pull-requests that are helpful in determining the activity of a project.

The proposal is to augment the existing method of collecting repository metrics with git hosting provider specific APIs providing additional metrics.

- [x] GitHub ( #161  )
- [ ] GitLab

imphil:
Let's close this for now as github is implemented and working. We can revisit gitlab support if there's a need for it from our users.

PR #161, opened 2017-07-27
Import additional metrics from GitHub

agathver:
Alternative to performing a `git clone` locally. Will also pull issues, forks, and pull requests.

- [x] Fetch README
- [x] Fetch License
- [x] Fetch stars, issues, pull requests and forks
- [x] Display github metrics

#154


imphil:
Thanks for your patience. I've had a look at the code and I'd like to propose a change in class structure here. Right now, you add things like "stars" and "forks" or number of issues to the source repository. However, that's not features of a repository itself, that are features of the source hosting site.

Consider the following cases:
- Project fully using github
- Project with bare git repo, issues tracked in bugzilla (many older projects)
- Project using github for the code, but phabricator for issues and reviews (lowrisc being an example here: http://phab.lowrisc.org/)

So, while we don't need or want any of these combinations right now, we should at least make sure that our code is structured in a way to allow extensions towards these use cases.

What does this mean?
- SourceRepo is representing a source repository, not the hosting site around it.
- Issues are a clear example of being orthogonal to it (e.g. IssueTracker <- GithubIssueTracker, etc.)
- Forks, stars/likes, etc. are a bit more tricky, but essentially that's a third category of data, essentially "data associated with the project, but on another project hosting site" I'm not sure how to exactly model that, maybe you have a proposal?

agathver:
@imphil 
>However, that's not features of a repository itself, that are features of the source hosting site.

Should we move them to the project entity ?

>So, while we don't need or want any of these combinations right now, we should at least make sure that our code is structured in a way to allow extensions towards these use cases.

We can add additional services to query the issue tracker (phabricator/github/gitlab/bugzilla) and fetch that data and store it in the entity. Since we are only concerned with counts and dates of the latest issue, the present model should be fine. They are just scalars.

>SourceRepo is representing a source repository, not the hosting site around it.

We can treat, issues, pull requests as properties of the "SourceRepo" or "Project".

>Issues are a clear example of being orthogonal to it (e.g. IssueTracker <- GithubIssueTracker, etc.)

I could not understand what have you stated here.  Are you proposing a form of inheritance using classes ?

imphil:
> We can treat, issues, pull requests as properties of the "SourceRepo" or "Project".

Or, option 3, create a new entity named something like "HostingSite", which has a 1:1 relationship with Project. Whether that makes sense greatly depends on the information we want to store. Currently, I recall:
- number of forks
- number of stars
- number of pull requests

I agree that for these three scalar values, going the extra mile of an additional entity does not seem worth the effort. Putting them into the SourceRepo is fine from my side. (Unless you plan to add more elaborate metrics in this regard, then we should revisit this.)

>> Issues are a clear example of being orthogonal to it (e.g. IssueTracker <- GithubIssueTracker, etc.)
>
> I could not understand what have you stated here. Are you proposing a form of inheritance using classes ?

Currently we have the Project entity, which has a 1:1 relationship with SourceRepo. In addition, Project has a field issueTrackerUrl, essentially representing the issue tracker. Now that we want to get more information about the issues than just the URL, the path forward is to replace issueTrackerUrl with a IssueTracker class, again with a 1:1 relationship with Project. There we can store the issueTrackerUrl, and other information, such as the number of open issues, etc. For specific issue tracker implementations, like the one from GitHub, IssueTracker can then be subclassed into GithubIssueTracker (just like it's done for the SourceRepo).

site/composer.json
```diff
@@ -34,7 +34,7 @@
 		"exercise/htmlpurifier-bundle" : "*",
 		"doctrine/doctrine-migrations-bundle" : "^1.0",
 		"symfony/http-kernel" : "~3.2.3",
-		"knplabs/github-api": "^2.3",
+		"knplabs/github-api": "dev-master",
```
oleg-nenashev: Does it mean "latest"? If yes, it may cause instability when a new release comes out

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -181,15 +181,15 @@ protected function commitExists(string $commitId): bool
...
+        $this->logger->info('Checking commits in ' . $cwd);
 
         $process = $this->processCreator->createProcess('git',
-                                                        [
-                                                            'merge-base',
-                                                            '--is-ancestor',
-                                                            $commitId,
-                                                            'HEAD',
-                                                        ]);
+            [
```
oleg-nenashev: unrelated refactoring

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GithubRepoCrawler.php
```diff
@@ -0,0 +1,174 @@
...
+        $variables = [
+            'owner' => $this->githubUser,
+            'repository' => $this->githubRepoName,
+        ];
+
+        $client = $this->getGithubClient();
+
+        if (null === $client) {
+            // we cannot extract any data from github in this case
+            $this->logger->error('Unable to fetch data from Github: No access token');
```
oleg-nenashev: It makes a presumption that this is an only cause of null client. It may change in the future. In languages like Java I would rather recommend throwing exceptions, not sure about PHP

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GithubRepoCrawler.php
```diff
@@ -0,0 +1,174 @@
...
+  $repository: String!
+ ) {
+  repository(owner: $owner, name: $repository) { 
+    forks {
+      totalCount
+    }
+    hasIssuesEnabled
+    updatedAt
+    pushedAt
+    issues(states:OPEN) {
```
oleg-nenashev: Does it include pull requests in this API version?

site/src/Librecores/ProjectRepoBundle/Util/MarkupToHtmlConverter.php
```diff
@@ -90,7 +90,7 @@ protected function markupToUnsafeHtml($markupInput)
                 throw new \RuntimeException("Unable to write to file $tmpFilename");
             }
 
-            $unsafeHtml = $this->markupFileToHtml($tmpFilename);
+            $unsafeHtml = $this->markupFileToUnsafeHtml($tmpFilename);
```
oleg-nenashev: lgtm, seems to be a bugfix in the previous code

site/composer.json
agathver: Yes, but it stores a commit sha in composer.lock, so everyone will get the same version.
We need some features (GraphQL GA) and it has not been released yet. It's mostly a temporary solution, we will pin a version number in composer.json, once the v2.6 is released.

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GithubRepoCrawler.php
agathver: Yes

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GithubRepoCrawler.php
agathver: I will change this to throw an exception.

site/src/Librecores/ProjectRepoBundle/Entity/SourceRepo.php
```diff
@@ -102,6 +102,46 @@
...
+     */
+    private $stars;
+
+    /**
+     * Watchers of this repository in the repository host
+     * @var int
+     *
+     * @ORM\Column(type="integer", nullable=true)
+     */
+    private $watchers;
```
agathver: GitHub blurs the distinction between a project and a repository. We can still consider it to a project

site/src/Librecores/ProjectRepoBundle/Util/GithubApiService.php
```diff
@@ -122,6 +127,23 @@ public function getAuthenticatedClient()
...
+            $oauthAccessToken = $user->getGithubOAuthAccessToken();
+            $client->authenticate($oauthAccessToken, null,
+                                  Github\Client::AUTH_URL_TOKEN);
+
+            return $client;
+        }
+
+        return null;
+    }
+
```
agathver: Fixed it

agathver:
I have made changes as pointed out.

>Don't overload the SourceRepoCrawler with things which don't have anything to do with the Repository, but with other feature of GitHub. Introduce a separate crawler instead.

The GitHub repo crawler only updates GitHub specific data, it mostly updates the project entity now.

agathver:
@imphil Fixed all nits you pointed out and split many long lines into smaller ones.

site/composer.json
oleg-nenashev: Ok, fine for me while it's not in production

Issue #163, opened 2017-08-03
Error: Could not convert database value "" to Doctrine Type array


imphil:
We get the following error when requesting https://www.stage.librecores.org/project/list on staging:

```
[2017-08-03 11:56:46] request.INFO: Matched route "librecores_project_repo_project_list". {"route":"librecores_project_repo_project_list","route_parameters":{"_controller":"Librecores\\ProjectRepoBundle\\Controller\\ProjectController::listAction","_route":"librecores_project_repo_project_list"},"request_uri":"https://www.stage.librecores.org/project/list","method":"GET"} []
[2017-08-03 11:56:46] security.INFO: Populated the TokenStorage with an anonymous Token. [] []
[2017-08-03 11:56:46] request.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\Types\ConversionException: "Could not convert database value "" to Doctrine Type array" at /var/www/lc/site/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ConversionException.php line 46 {"exception":"[object] (Doctrine\\DBAL\\Types\\ConversionException(code: 0): Could not convert database value \"\" to Doctrine Type array at /var/www/lc/site/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ConversionException.php:46)"} []
```

It seems a database migration is missing/wrong?

imphil:
Switched staging to dev mode, a full stack trace is now available at the given URL.

imphil:
```
[1] Doctrine\DBAL\Types\ConversionException: Could not convert database value "" to Doctrine Type array
    at n/a
        in /var/www/lc/site/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ConversionException.php line 46

    at Doctrine\DBAL\Types\ConversionException::conversionFailed('', 'array')
        in /var/www/lc/site/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ArrayType.php line 60

    at Doctrine\DBAL\Types\ArrayType->convertToPHPValue('', object(MySQL57Platform))
        in /var/www/lc/site/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php line 297

[...]

PR #164, opened 2017-08-06
Add default value to languageStats column in DB

agathver:
Fixes: #163 Error: Could not convert database value "" to Doctrine Type array

Manually inserts `'a:0:{}'` to newly created column, this is the PHP serialized representation of an empty array.


imphil:
Changing the migration actually doesn't fix the existing deployments (like on staging). However, let's take this for production and I'll fix staging manually.

oleg-nenashev:
LGTM, thanks for the fix

PR #167, opened 2017-08-17
Extract releases from Git tags

agathver:
This PR extracts releases from git tags, resembling a version tag.

Based on some patterns in release tags widely used in many repositories. Though, it may not be very accurate, it is very similar to how many services (such as composer and npm) count releases.

Examples of detected tags:
- v1.x.x-desc (semver)
- version1.x-something
- platform-v1.x.x (used in https://github.com/openrisc/mor1kx)
- text-v2.x-RC2
- text-v1.2-text
- artefact-1.23
- 1.2.4-alpha
- 1.2.45
- 1.2


olofk:
What about x.y or x.y.z tags? I use that for FuseSoC, but maybe it's not that common

agathver:
@olofk Most repositories I saw used tags in this format. However, it isn't difficult to support x.y tags too.
Still a having a prefix is the recommended way.

olofk:
I guess you're right. I should probably retag all the releases with a v prefix some day

oleg-nenashev:
> I guess you're right. I should probably retag all the releases with a v prefix some day

I would say that releases without "v" prefix are also very popular. People also use common Maven/Gradle flows (even for hardware), and I would advice supporting the `${artifactID}-${version}` format (`version` - everything you mentioned above, usually without "v" :( ). 

But it can be added on-demand, no need to do everything right now

agathver:
@oleg-nenashev Supporting the `${artifactID}-${version}` is a matter of adding a regex. I've added a commit to detect them.

site/src/Librecores/ProjectRepoBundle/Entity/ProjectRelease.php
```diff
@@ -0,0 +1,130 @@
...
+     * Is the release a pre-release
+     *
+     * @var bool
+     */
+    private $isPrerelease;
+
+    /**
+     * Commit ID this release is based on
+     *
+     * @var string
```
imphil: The idea is to use this class also for releases which are not derived from the source repo, yes? In this case, commit id might not be available. Make it nullable.

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -429,6 +451,62 @@ private function parseCommits(string $outputString) : int
     }
 
     /**
+     * Update project releases
+     *
+     * @return bool if successful
+     */
+    protected function updateReleases() {
```
imphil: nit: { into next line

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -429,6 +451,62 @@ private function parseCommits(string $outputString) : int
...
+
+        try {
+            $this->mustExecuteProcess($process);
+            $output = $process->getOutput();
+        } catch(\Exception $e) {
+            $this->logger->error("Unable to fetch releases from $cwd" .
+             'Process execuion failed: ' . $e->getMessage());
+            return false;
+        }
+            $tagRegex = '/^([\w-]*v?.*\d+\.\d+.*)\|([[:xdigit:]]+)\|(.+)$/i';
```
imphil: Seems like the indentation broke here.

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -429,6 +451,62 @@ private function parseCommits(string $outputString) : int
...
+            $preReleaseRegex = '/.*[-_\.](alpha|beta|RC-?\d+).*/i';
+
+            $lines = explode("\n", trim($output));
+            $releases = [];
+
+            foreach($lines as $line) {
```
imphil: nit: `foreach ($lines`

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -429,6 +451,62 @@ private function parseCommits(string $outputString) : int
...
+                try {
```
imphil: indentation again

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -335,7 +338,7 @@ protected function getDescriptionSafeHtml(): ?string
         }
 
         return $sanitizedHtml;
-    }
+    }This is not an entity, rather and array, so variables are by-default nullable.
```
imphil: What went wrong here? (and on the ones above)?

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -449,6 +452,64 @@ private function parseCommits(string $outputString) : int
...
+                if (preg_match($tagRegex, $line, $matches)) {
+                    $release = new ProjectRelease();
+                    $release->setName($matches[1])
+                            ->setPublishedAt(new \DateTime($matches[3]))
+                            ->setCommitID($matches[2])
+                            ->setIsPrerelease(preg_match($preReleaseRegex, $matches[1]));
+                    $releases[] = $release;
+                }
+            } catch (\Exception $e) {
+                $this->logger->warning("Skipped due to parse error: $line");
```
imphil: *What* was skipped due to parsing errors?

PR #168, opened 2017-08-21
Add SiFive blog to planet

olofk:


imphil:
Yes, it's broken until we get to include the Docker image into travis and run the tests there. That doesn't prevent that PR from being merged, though :-) Thanks Olof.

PR #170, opened 2017-08-23
Enable xdebug remote debugging and add phpMyAdmin

agathver:
Adds phpMyAdmin and [xdebug remote debugging](https://xdebug.org/docs/remote) to ansible. Only installed in Vagrant.

phpMyAdmin can be accessed through http://pma.librecores.devel

### DIFF
--- a/ansible/roles/dev-vagrant/handlers/main.yml
+++ b/ansible/roles/dev-vagrant/handlers/main.yml
@@ -1,0 +1,5 @@
+- name:    restart php-fpm
+  service:
+    name: "php{{ php_version }}-fpm"
+    enabled: yes
+    state: restarted
--- a/ansible/roles/dev-vagrant/handlers/main.yml
+++ b/ansible/roles/dev-vagrant/handlers/main.yml
@@ -3,3 +3,9 @@
     name: "php{{ php_version }}-fpm"
     enabled: yes
     state: restarted
+
+- name: restart nginx
+  service:
+    name: nginx
+    enabled: yes
+    state: restarted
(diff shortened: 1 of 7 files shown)


imphil:
- Please add a note on the developer documentation about xdebug remote debugging.
- On PMA: Isn't the easyadmin bundle which we already integrate in Vagrant doing the job? PMA is kind of the ugly kid on the block from the last PHP decade; however, if its something which makes you productive, I'm not against it -- let's just make sure that the existing tools (easyadmin bundle) are not sufficient.

agathver:
>Isn't the easyadmin bundle which we already integrate in Vagrant doing the job?

It allows to take dumps, run arbitrary SQL queries easily.

>let's just make sure that the existing tools (easyadmin bundle) are not sufficient.

PMA is made to interact with a database, EasyAdmin is geared for creating a dashboard like application with integrity enforcement etc. It is for editing data in production. It doesn't allow to drop all rows without reference checking or inspect raw database values that easily, which are needed while testing migrations.

On a side note: May be EasyAdmin should be added to the production version to allow user account moderation, project reassignment or other admin level tasks.

imphil:
OK. Then please add the docs for xdebug and where to find PMA and this one is ready to go.

imphil:
@agathver while writing the missing documentation for this PR, i saw that the commits are missing the file 99-librecores.ini, the xdebug template. Could you add that and then I'll add the documentation for it?

imphil:
Ah, never mind, I just missed the file yesterday. All good @agathver !

agathver:
@imphil Currently busy with exams will make changes after a week.

imphil:
@agathver I already included the documentation changes and merged the PR. Thanks!

PR #176, opened 2017-09-06
Add ZipCPU to planet

olofk:


imphil:
Seems no https is available since it's served from gh pages and using CNAME to redirect to a custom domain :-( Which means pictures will not be available, but that's expected.

PR #181, opened 2017-09-22
Fix bug in project quality stars display

### LINKED ISSUE #177: 5/5 stars shows 7 stars
An image is better than thousand words.

grafik

as seen on https://www.librecores.org/olofk/FuseSoC

agathver:
Closes #177

No more 7 stars!

### DIFF
--- a/site/src/Librecores/ProjectRepoBundle/Resources/views/Project/view.html.twig
+++ b/site/src/Librecores/ProjectRepoBundle/Resources/views/Project/view.html.twig
@@ -37,9 +37,12 @@
           {% if metadata.qualityScore.halfStars == 1 %}
             <span class="fa fa-star-half-full" aria-hidden="true"></span>
           {% endif %}
-          {% for _ in range(metadata.qualityScore.fullStars + metadata.qualityScore.halfStars + 1, 5) %}
-            <span class="fa fa-star-o" aria-hidden="true"></span>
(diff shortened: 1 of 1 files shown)


site/src/Librecores/ProjectRepoBundle/Resources/views/Project/view.html.twig
```diff
@@ -37,9 +37,12 @@
           {% if metadata.qualityScore.halfStars == 1 %}
             <span class="fa fa-star-half-full" aria-hidden="true"></span>
           {% endif %}
-          {% for _ in range(metadata.qualityScore.fullStars + metadata.qualityScore.halfStars + 1, 5) %}
-            <span class="fa fa-star-o" aria-hidden="true"></span>
-          {% endfor %}
+          {% set stars = metadata.qualityScore.fullStars + metadata.qualityScore.halfStars %}
```
oleg-nenashev: Handling of half-stars is still a bit suspicious. If we somehow get 3 half-stars in the input, the output will be messed up in an interesting way. but it fixes the bug for sure

agathver: Half stars can never be 3, they can either be 0 or 1. There is a `% 2` in the code for calculating half stars

PR #182, opened 2017-09-23
Fix error when listing organitations

wallento:
Symfony was giving an error when a user listed repositories:

    Type error: Argument 1 passed to Librecores\ProjectRepoBundle\Repository\OrganizationRepository::findAllByMemberOrderedByName() must be an instance of Librecores\ProjectRepoBundle\Repository\User, instance of Librecores\ProjectRepoBundle\Entity\User given, called in /var/www/lc/site/src/Librecores/ProjectRepoBundle/Controller/OrganizationController.php on line 25

This fix is rather trivial by importing the correct class into the
Repository.


eliaskousk:
I'm still rebuilding the VM to test but I bet that this missing include fixes the issue.

PR #183, opened 2017-09-23
Visually improve the organization view

wallento:
I find the organization view somewhat hard to scan as it seems pretty
flat. Hence I propose these improvements:

 * Visually highlight the "Name:", .. field names

 * Re-arrange the ordering and put supporters up between projects and
   users

 * Put the join actions right next below the members

 * Put the actions "Settings", "Leave", "Delete" in a new category
   "Actions"

 * Add rulers between the categories

I hope this helps improve it.

Finally, I changed some html coding: I think headings are not allowed
to be inside the paragraphs. It at least seems counterintuitive to me,
but I would even think its formally not valid.


wallento:
True. I was thinking to add optional stuff lower on the page (if admin), but you are right about importance of information. I will update that. Beyond that, as you say, its just a marginal improvement, I may amend this one until @imphil is back.

imphil:
Thanks @wallento Probably not the last stab at this, but certainly an improvement.

Issue #185, opened 2017-09-24
Using Sentry for tracking errors in production


agathver:
[Sentry](https://sentry.io) is an error tracking tool, that collects errors in a dashboard. 

Since, we do not have enough tests and unexpected things happen in production, it takes time to discover those hidden bugs. It will help us to track them down.

It's ridiculously easy to setup and integrate with Symfony. Integrates with Github to report issues automatically. Supports separate environments like dev,staging,production. Also has a self-hosted version too.

wallento:
Excellent initiative! I think it addresses a very important issue. Can we kick it off by integrating it in a PoC? I have created an account and suppose the credential ("DSN") is something to be kept secret. We would need to write it to `app/config/config.xml` using ansible, right?

wallento:
Okay, I had a look at it. Its so simple, even I could do it, hence assigned to me

agathver:
>I have created an account and suppose the credential ("DSN") is something to be kept secret.

A part of the key may be embedded in client side JavaScript for frontend error tracking too. But since we do not use much JS yet, we can skip configuring that.

>We would need to write it to app/config/config.xml using ansible, right?

Yes, but only on prod setups. Leave the DSN key defined as a blank value on dev setups, which disables Sentry. Else, all kinds of errors will flood the notification channel!

wallento:
Thanks, I found there is a public DSN for browser stuff. Looks good, very helpful tool!

imphil:
Deployed to production and staging

PR #186, opened 2017-09-24
Enable sentry crash reports

wallento:
@agathver proposed to use http://sentry.io to track unhandled issues on LibreCores.org. It makes a lot of sense to use it, even pay the money for a better subscription. This adds it to LibreCores (very simple). Just set the `sentry_dsn` entry for the production and the staging environment. For dev environments it should be empty, errors are properly reported there.

### DIFF
--- a/site/composer.json
+++ b/site/composer.json
@@ -38,7 +38,8 @@
 		"php-http/guzzle6-adapter": "^1.1",
 		"symfony/cache": "^3.2",
 		"javiereguiluz/easyadmin-bundle": "^1.16",
-		"beberlei/DoctrineExtensions": "^1.0"
+		"beberlei/DoctrineExtensions": "^1.0",
+		"sentry/sentry-symfony": "^0.8"
 	},
 	"scripts" : {
 		"post-install-cmd" : [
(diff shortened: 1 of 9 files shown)


wallento:
Fixes #185

site/composer.json
```diff
@@ -38,7 +38,8 @@
 		"php-http/guzzle6-adapter": "^1.1",
 		"symfony/cache": "^3.2",
 		"javiereguiluz/easyadmin-bundle": "^1.16",
-		"beberlei/DoctrineExtensions": "^1.0"
+		"beberlei/DoctrineExtensions": "^1.0",
+		"sentry/sentry-symfony": "dev-master"
```
agathver: I think we should stick with the 0.8.x series for now. 1.0 may have some issues

agathver:
I'll pull this and test it locally and see what is causing the timeouts

site/composer.json
wallento: thanks, I will rebase with the correct version.

agathver:
If the communication between sentry servers is the issue, we can always use an on-premise sentry installation. They provide a convenient Dockerfile.

wallento:
Yeah, but I think it will not be an issue once it runs on staging or production, because both are AWS and not VirtualBox and my crappy dialup.

Issue #192, opened 2017-09-28
Password resets


agathver:
We would like to have proper password reset emails. We use [FOSUserBundle which provides logic and starter templates](http://symfony.com/doc/master/bundles/FOSUserBundle/emails.html) for password reset.

However, the UI needs to match the design of LibreCores.org

eliaskousk:
I have been assigned the relevant FOS user bundle improvements
(registrations, lost password etc) a couple of months or more ago but
unfortunately I still haven't found time to complete them. I'll try harder
to find the time so give me until next month and if I don't deliver by all
means let's assign this to someone else!

wallento:
Hi @eliaskousk,

the idea is to have a few simpler sub-tasks for hacktoberfest. Do you think this is not separable from your work?

Cheers,
Stefan

eliaskousk:
This should be simple enough to get down over this weekend. I'll take a
look and let you know.

agathver:
Hi @eliaskousk 
>I have been assigned the relevant FOS user bundle improvements (registrations, lost password etc) a couple of months or more ago but unfortunately I still haven't found time to complete them

Oh! I wasn't aware of that.

The aim here is to put up simple tasks so that we can get new contributors. Since we know the way around the source code, we should concentrate on more complex tasks.

eliaskousk:
Hi @agathver,

since you recent very good work I think I belong to this category haha :)
Seriously though I'll take a look at it and see if I can handle it or not (due to me being terrible on frontend work).

scarecrow21:
Hi. I'm a newbie here. Will someone guide me through solving this?

sandipbhuyan:
Is it ope?. I want to work on this.

agathver:
// cc @eliaskousk

imphil:
@sandipbhuyan this is mostly resolved in PR https://github.com/librecores/librecores-web/pull/193. However, in order to merge this we first need a verified e-mail address from the user. Would that be something for you to work on? There is more discussion on that in https://github.com/librecores/librecores-web/issues/142

eliaskousk:
@imphil Is the plan for the AWS e-mail service still on? Let me know if I can somehow help with this. I'll take a look at the other issues as well and see if there is anything urgent.

imphil:
@eliaskousk yes, the plan is to use AWS email for such transactional mail. Probably the most simple route is simply setting up swiftmail using their SMTP servers

sandipbhuyan:
[...]

Issue #142, opened 2017-05-29
Verify user email addresses


imphil:
Verify email addresses given by users on registration and when changing the email address by sending them a link to click on.

Users can register and use the site without being verified, however, no mails should be sent to email addresses which are not verified (except for the verification mail itself, of course).

Dependency: https://github.com/librecores/librecores-web/issues/94

eliaskousk:
Thanks for opening the issue, you were quicker. This feature should be much easier than the complete notification system (#143) and thus I will begin working on this. I will try to make it a bit abstract so that the actual generation of e-mails can be reused in #143.

Agathver:
Isn't verification of user emails a standard part of FOSUserBundle ?

eliaskousk:
Yes the FOS bundle supports verification of user emails during registration but we currently don't use it and haven't tested it on either dev or production. For the latter we will use an AWS email service that Philipp will setup soon (possibly in parallel).

Also changing email address requires some attention and we also have the option of not allowing it.

Last but not least, thank you for the html emails repo, this should be great for our purposes since I agree that we don't want to mess with handcrafting emails without some help by such template library.

PR #193, opened 2017-10-08
Password Reset

eliaskousk:
This PR adds password reset functionality from the FOS user bundle as discussed in #192.

The link was added to the login page and the reset password UI now matches the reset of the site.

The four templates that are used for resetting the password are the following and correspond to the steps that the user follows when starting the reset process:

1. To reset your password visit the route with path `/user/request` which contains a simple form to fill your email address or username. The template for this page can be found at `site/app/Resources/FOSUserBundle/views/Resetting/request.html.twig`
2. A message is shown after you submit an existing email / username on the previous form. The url of this message page is `/user/check-email?username=$USERNAME$` and the template can be found at `site/app/Resources/FOSUserBundle/views/Resetting/check_email.html.twig`
3. The email that gets sent to the user with the token link can be found at `site/app/Resources/FOSUserBundle/views/Resetting/email.html.twig`
4. The link that you must follow on your email has an url of the form `/user/reset/$TOKEN$` and the template can be found at `site/app/Resources/FOSUserBundle/views/Resetting/reset.html.twig`

The default email template on step No 3 is still the default, i.e a very simple message. We should change this to something in proper English and possibly devise a common email template to also use for other email purposes like e.g registration (we currently don't send anything) as discussed in #142. I just saw that @agathver has already proposed that we use https://github.com/rodriguezcommaj/salted for the email templates so let's see what it can offer.

Last but not least I have disabled the default FOS user profile routes because the `user/profile/show` one is redundant (we already have a profile settings page) and more importantly it allows the user to access the `user/profile/edit` page where currently he/she can edit the username or email completely uncontrolled.

### DIFF
--- a/site/app/Resources/FOSUserBundle/views/Resetting/check_email.html.twig
+++ b/site/app/Resources/FOSUserBundle/views/Resetting/check_email.html.twig
@@ -1,0 +1,17 @@
+{% extends "@FOSUser/layout.html.twig" %}
+
+{% trans_default_domain 'FOSUserBundle' %}
+
+{% block fos_user_content %}
+    <div class="center-block">
+        <div class="panel panel-default">
+            <div class="panel-heading"><h3 class="panel-title">Reset Password</h3></div>
+            <div class="panel-body">
+                <p>
+                    {{ 'resetting.check_email'|trans({'%tokenLifetime%': tokenLifetime})|nl2br }}
+                </p>
+            </div>
+        </div>
+    </div>
+    </div>
+{% endblock fos_user_content %}
--- a/site/app/Resources/FOSUserBundle/views/Resetting/check_email.html.twig
+++ b/site/app/Resources/FOSUserBundle/views/Resetting/check_email.html.twig
@@ -3,15 +3,14 @@
 {% trans_default_domain 'FOSUserBundle' %}
 
 {% block fos_user_content %}
-    <div class="center-block">
-        <div class="panel panel-default">
-            <div class="panel-heading"><h3 class="panel-title">Reset Password</h3></div>
-            <div class="panel-body">
-                <p>
-                    {{ 'resetting.check_email'|trans({'%tokenLifetime%': tokenLifetime})|nl2br }}
-                </p>
-            </div>
-        </div>
+  <div class="center-block">
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3 class="panel-title">Reset Password</h3>
+      </div>
+      <div class="panel-body">
+        <p>{{ 'resetting.check_email'|trans({'%tokenLifetime%': tokenLifetime})|nl2br }}</p>
+      </div>
     </div>
-    </div>
+  </div>
 {% endblock fos_user_content %}
--- a/site/app/Resources/FOSUserBundle/views/Resetting/email.txt.twig
+++ b/site/app/Resources/FOSUserBundle/views/Resetting/email.txt.twig
@@ -1,0 +1,13 @@
+{% trans_default_domain 'FOSUserBundle' %}
+{% block subject %}
+{%- autoescape false -%}
+{{ 'resetting.email.subject'|trans({'%username%': user.username}) }}
+{%- endautoescape -%}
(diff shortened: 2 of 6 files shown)


imphil:
Thanks for this PR @eliaskousk 

Before we can get this merged, we have another problem to solve first: Password reset functionality requires verified email addresses ( #142 ) to avoid creating a security hole. Since we import unverified (at least unverified on LibreCores) email addresses when logging in through GitHub, Google and other OAuth providers, the password reset mechanism must also check a field in the DB if an email address is actually verified.

On the code: it looks roughly fine to me, except please indent HTML files with two spaces to keep it consistent with our other code.

agathver:
>Password reset functionality requires verified email addresses ( #142 ) to avoid creating a security hole.

Perhaps it is a necessary security hole. What if the user who wants to verify his email forgot the password? I have not seen many sites making this check.

eliaskousk:
Hey @imphil, I can look at this again during the weekend.
I'll let you know how it goes and if I need any help.

eliaskousk:
@imphil I have add the html body for the email template and indented twig code with 2 spaces.

> the password reset mechanism must also check a field in the DB if an email address is actually verified.

Should this be a check on the existence of the OAuth fields being not null ( i.e check for githubOAuthUserId or googleOAuthUserId) which implies that the user didn't create a local account and thus his e-mail is potentially unverified (an impossible scenario but I understand that as far as we know these email addresses are unverified by us).

And what about those that registered with a local account (not OAuth) and didn't verify their email since we didn't have the e-mail verification when we launched.

So to implement what you suggest I should better add a new db field on the user entity, i.e `email_verified` which will be null / 0 for all existing users and check this when someone requests a password reset regardless of their account type (local / OAuth).

imphil:
> Should this be a check on the existence of the OAuth fields being not null ( i.e check for githubOAuthUserId or googleOAuthUserId) which implies that the user didn't create a local account and thus his e-mail is potentially unverified (an impossible scenario but I understand that as far as we know these email addresses are unverified by us).

In the current state it's sufficient to check for the "enabled" flag in the account and treat accounts with enabled == false as unverified.

In PR #142 I've come to the conclusion that we can trust the email addresses we get from our current OAuth providers, GitHub and Google, and don't need to create a more involved email verification scheme as I initially thought. GitHub doesn't allow OAuth without a verified email, and the Google email address is verified by Google as well -- two entities which should know what they're doing, at least better than we do.

If we trust those two, we're back to the default FOSUserBundle flow and can just use the enabled field as verified indicator.

> And what about those that registered with a local account (not OAuth) and didn't verify their email since we didn't have the e-mail verification when we launched.

I'll see how we can get these email addresses verified, most likely using a manual approach for the (rather small) amount.

Given all that I think your patch set should be already sufficient, I'll look through the code in more detail in a bit.

imphil:
I've now tested the changes and looked at the code. It all looks good, except for one rather odd thing: resetting a password enables an account. Try the following:
- Create a new account on Librecores (not using OAuth). *Do not* confirm the email address.
- Click on reset password
- Click on the link in the reset password email
- Now you're able to log into LibreCores without having clicked on the "confirm email" link.

This is a known bug in FOSUserBundle (https://github.com/FriendsOfSymfony/FOSUserBundle/issues/2298 and https://github.com/FriendsOfSymfony/FOSUserBundle/issues/2598, but given the less than optimal maintenance of that bundle not yet resolved.

However, I don't think it's worth working around that: in the end, the user clicked on a link in an email we sent him/her. I could imagine a couple scenarios in which this is less secure than the expected behavior (don't enable users through the password reset), but for now I'm fine with merging this just as-is.

@eliaskousk your call. Do you want to see if you can work around the bug in FOSUserBundle, or just we just leave it as-is?

eliaskousk:
@imphil Still haven't time to work around the bug but I haven't forgotten it. I suppose we can wait another couple of weeks and if I still can't find time we can merge and resolve it in the future (or not!).

imphil:
This is now merged as-is, accepting the outlined problem for now.

Issue #195, opened 2017-10-13
Update service definitions to use new service IDs


imphil:
To make sure our code base stays up to date with best practices, we should switch our Symfony services to use the currently available features.

Follow these guides:
- https://knpuniversity.com/screencast/symfony-3.3/class-ids
- https://knpuniversity.com/screencast/symfony-3.3/private-services
- https://knpuniversity.com/screencast/symfony-3.3/auto-register-services
- https://knpuniversity.com/screencast/symfony-3.3/controllers-services

and subsequent chapters

annika1:
Hey Philipp,
I would like to work on this issue. I guess the best idea is to go though the tutorial to symphony3.3 you have linked above?

imphil:
Annika, great to see you join! You first need to set up a development environment, however. That's documented here: http://librecores-web.readthedocs.io/en/latest/devenv/quickstart.html  Let me know if you face any problems during this first step.

As next steps, as you said, follow the tutorial linked above (and just move on to the next chapters). It's probably best to start with the first (class-ids) topic/chapter and create a PR to get feedback early. Then you can move on to the next ones and update the PR iteratively.

annika1:
I have already set up the development environment and can access the development page http://www.librecores.devel, therefore I will start now with the tutorial and the chapter to class-ids. I let you know, when problems come up.

agathver:
Fixed in #312, #340

PR #204, opened 2017-11-29
Use new symfony 3.3 features for services

annika1:
For updating the librecores-web code to Symphony 3.3 I've stated to work on the issue #195. This pull request includes only the first changes to the code regarding the services.


site/app/config/services.yml
```diff
@@ -4,8 +4,12 @@ parameters:
...
+        public: false
+
     # Twig date extension - http://twig-extensions.readthedocs.io/en/latest/date.html
-    twig.extension.date:
-       class: Twig_Extensions_Extension_Date
-       tags:
-            - { name: twig.extension }
+    twig.extension.date: # html template
+        class: Twig_Extensions_Extension_Date # getagged class
+
```
imphil: Your changes broke the extension: http://www.librecores.devel/project/list is showing " Updated diff.ago.month " only. "diff.ago.month" is the translation string, i.e. we're missing the translations. This behavior is caused by autowire injecting a TranslatorInterface into the constructor of the extension. If that happens, the extension thinks it can be translated -- but we don't have any translations available. Adding `autowire: false` solves this problem.

Additionally, since version 1.5.0 of the twig extensions namespaced classes were added, so we can now make use of the classname=servicename paradigm that you're using below.

```patch
diff --git a/site/app/config/services.yml b/site/app/config/services.yml
index 0c3d3e0..c667729 100644
--- a/site/app/config/services.yml
+++ b/site/app/config/services.yml
@@ -10,6 +10,9 @@ services:
         public: false
 
     # Twig date extension - http://twig-extensions.readthedocs.io/en/latest/date.html
-    twig.extension.date: # html template
-        class: Twig_Extensions_Extension_Date # getagged class
-
+    Twig\Extensions\DateExtension:
+        # XXX: autowiring is disabled to make sure no TranslatorInterface is
+        # passed to the constructor. If it is, we need to add translations
+        # for date.ago.xxx keys as discussed here:
+        # http://twig-extensions.readthedocs.io/en/latest/date.html#translation
+        autowire: false
diff --git a/site/composer.json b/site/composer.json
index adb0e92..21a706a 100644
--- a/site/composer.json
+++ b/site/composer.json
@@ -21,7 +21,7 @@
                "doctrine/dbal" : "~2.6.0",
                "doctrine/doctrine-bundle" : "~1.7.0",
                "doctrine/doctrine-fixtures-bundle" : "<2.5",
-               "twig/extensions" : "~1.0",
+               "twig/extensions" : ">=1.5.0",
                "sensio/distribution-bundle" : "~5.0.18",
                "sensio/framework-extra-bundle" : "~4.0.0",
                "incenteev/composer-parameter-handler" : "~2.0",
```

(You need to run `cd /var/www/lc/site; ./bin/console cache:clear` inside the VM to see the changes.)

site/src/Librecores/ProjectRepoBundle/Resources/config/services.yml
```diff
@@ -1,89 +1,92 @@
...
+
+
 services:
     _defaults:
         autowire: true
         autoconfigure: true
+        public: false
+
+    # makes classes in src/AppBundle available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
```
imphil: Please remove the comment (or change AppBundle to the class name that's used)

site/src/Librecores/ProjectRepoBundle/Resources/config/services.yml
```diff
@@ -1,89 +1,92 @@
...
+    Librecores\ProjectRepoBundle\:
+        resource: '../../../ProjectRepoBundle/*'
+        # you can exclude directories or files
+        # but if a service is unused, it's removed anyway
+        exclude: '../../../ProjectRepoBundle/{Entity,Repository,DataFixtures}'
+        # TODO: Evaluate why there is a Problem with class inside the folder DataFixtures.
```
imphil: can you make this comment more specific? Other than that, excluding DataFixtures seems a reasonable thing to do.

annika1:
Hi Phillip,
I reset the old commit. Here is the new one with your remarks. I will further take a look on the following chapters in the Symfony 3.3 tutorial.

site/app/config/legacy_aliases.yml
```diff
@@ -0,0 +1,2 @@
+services:
+        twig.extension.date: '@Twig\Extensions\DateExtension'
```
imphil: That one should not be needed, we don't refer to the extension by name anywhere.

site/app/config/services.yml
```diff
@@ -3,9 +3,18 @@
 parameters:
 #    parameter_name: value
 
+imports:
+    - { resource: legacy_aliases.yml }
+
```
imphil: When removing the only content in this file above you can remove this import as well.

site/app/config/services.yml
```diff
@@ -3,9 +3,18 @@
...
+        autoconfigure: true
+        public: false
+
     # Twig date extension - http://twig-extensions.readthedocs.io/en/latest/date.html
-    twig.extension.date:
-       class: Twig_Extensions_Extension_Date
-       tags:
-            - { name: twig.extension }
+    #twig.extension.date: # html template
+    #    class: Twig_Extensions_Extension_Date # getagged class
```
imphil: Please remove this comment.

site/app/config/services.yml
```diff
@@ -3,9 +3,18 @@
...
+    Twig\Extensions\DateExtension:
+        autowire: false
```
imphil: ... and add back the comment I proposed (or a similar one) explaining why we have to set autowire to false in this case.

site/src/Librecores/ProjectRepoBundle/Resources/config/services.yml
```diff
@@ -1,89 +1,94 @@
...
+    # makes classes in src/Librecores/ProjectRepoBundle available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
+    Librecores\ProjectRepoBundle\:
+        resource: '../../../ProjectRepoBundle/*'
+        # you can exclude directories or files
+        # but if a service is unused, it's removed anyway
+        exclude: '../../../ProjectRepoBundle/{Entity,Repository,DataFixtures}'
+        # The class in the file Librecores\ProjectRepoBundle\DataFixtures\ORM\LoadDevTestData is named
+        # "LoadProjectData" and not as expected "LoadDevTestData", therefore it cannot be created a service for this
+        # class and is excluded above.
```
imphil: good catch! Please rename the class in LoadDevTestData.php to LoadDevTestData to make the autoloader happy.

site/src/Librecores/ProjectRepoBundle/Resources/config/services.yml
```diff
@@ -1,89 +1,94 @@
...
         autowire: true
         autoconfigure: true
+        public: false
+
+    # makes classes in src/Librecores/ProjectRepoBundle available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
+    Librecores\ProjectRepoBundle\:
+        resource: '../../../ProjectRepoBundle/*'
+        # you can exclude directories or files
+        # but if a service is unused, it's removed anyway
```
imphil: Please remove those comments.

imphil:
Annika, sorry for the delay. Github doesn't send notifications if pushes are added to an existing PR, and I missed this one. (To prevent this in the future, please add a small comment when your changes are done and the PR is ready to be re-reviewed or merged.)

This looks good now, merging. Thanks for your work!

Issue #210, opened 2018-03-01
Styling and Content of page after email is verified


imphil:
Currently we show the following screen after users click on the "verify email address" link in their inbox:

![grafik](https://user-images.githubusercontent.com/1467123/36872839-f2a8cb22-1da6-11e8-9f62-a9c120530007.png)

Can we
- make the styling a bit nicer so that the content isn't crammed into the top corner?
- add a bit more useful content, e.g. how to get started (add projects, browse through existing ones).

sandipbhuyan:
@imphil assign me this task. I want a demo layout of the content or a brief description of the content. I will be working on this after my next PR.

PR #216, opened 2018-03-06
Styling and Content of page after email is verified 

sandipbhuyan:
Resolving issue #210 
I have added one page named as confirmed.html.twig that overrides the default confirm pages.
I have added some svg files for the content of the page. 
The UI will be like
screen shot 2018-03-06 at 9 38 03 pm


imphil:
Thanks for the changes. I'll look through them over the weekend. How can I test the page? Do I need to create a new account or are there other ways? (http://www.librecores.devel/user/register/confirmed gives me an exception 'Attempted to call an undefined method named "getProviderKey" of class "HWI\Bundle\OAuthBundle\Security\Core\Authentication\Token\OAuthToken".').

imphil:
I'm logged in, but through GitHub (OAuth). This seems to be a bug/missing feature in HWIOAuth, but I didn't find time to look at it yet.

sandipbhuyan:
@agathver I am replacing the icons with the open source files. @imphil you can see this page when you will confirm your account. So you need to create a new account.

sandipbhuyan:
@agathver @imphil I have replaced the icons with open source icons and added the author information on README

imphil:
I've looked through it and I have some suggestions written inside the image itself.

librecores

sandipbhuyan:
@imphil I am modifying the view

sandipbhuyan:
Would I put this view on that page
screen shot 2018-03-12 at 1 56 18 am

imphil:
@sandipbhuyan sorry for the delay. Yes, you can make use of the front-page graphics, it already provides a rather actionable starting point. But if we're going this route, why not simply redirect to the front page? What additional value can we offer to people that have just confirmed their email?

Essentially, you don't need to be logged in to download or search for a core, you also don't need to have an account to access the gitter channel, etc. The most likely option why people chose to open an account is that they want to add a new project. Maybe we can highlight this option after the email confirmation finished?

Issue #178, opened 2017-09-14
Add help to "We have not analyzed this project yet"


wallento:
Hi,

it would be useful to show a small question mark next to this text too and maybe have a popup thingy explaining what it means and why.

Cheers

lakhanshiva:
I have started to work on this commit. I think you should take a look at the pull request i have created referencing this issue..

PR #228, opened 2018-04-06
Add help to- We have not analyzed this project yet, #178

lakhanshiva:
Commit to resolve the issue, #178 .
It has been tested locally. Please find a screenshot below.
issue_resolved_178


imphil:
Thanks @lakhanshiva for the PR. Unfortunately the information given in the tooltip is wrong: if we are still processing, the page is not even displayed. The "we have not analyzed this project yet" feature has different reasons, most likely not enough data being available. @agathver can you jump in with a more detailed explanation?

imphil:
Given the wider range of possible reasons, how about linking to a description page, as opposed to using a tooltip @lakhanshiva?

PR #241, opened 2018-06-06
Php coding style

### LINKED ISSUE #159: Add a linter based on projects code-style
To aid contributors and to lessen the burden while reviewing code, we should add a linter like [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer). It can be made as a part of the CI.

In the discussions at Gitter, we decided that we would follow [Symfony Coding Standards](https://symfony.com/doc/current/contributing/code/standards.html) with some modifications:
1. We will follow PSR-2 recommendations in regards to line length : 80 being the recommended length and 120 being a soft-limit.
2. We will not be strict about "yoda-conditions". Readability is the first priority.
3. We will not have strict restrictions for line-wrapping.
4. We will allow deviations from the standard for readability.

imphil:
This adds a phpcs configuration to the Symfony project and configures it with our style. I had to go through a lot of code (some manually, some is auto-fixed) to make sure we have a clean run as a base. I'm not particularly fond of the multi-line string splitting that the Symfony/PSR-2 style enforces, but it#s not ugly enough to care too much about it, so I've fixed the relevant occurrences to conform to the style.

Should fix #159

@agathver Could you have a look if you have time? With such an amount of changes it's easy to miss something.

### DIFF
--- a/site/app/AppKernel.php
+++ b/site/app/AppKernel.php
@@ -27,6 +27,7 @@ class AppKernel extends Kernel
             new Exercise\HTMLPurifierBundle\ExerciseHTMLPurifierBundle(),
             new Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle(),
             new Sentry\SentryBundle\SentryBundle(),
+            new Http\HttplugBundle\HttplugBundle(),
         );
 
         if (in_array($this->getEnvironment(), array('dev', 'test'))) {
--- a/site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php
+++ b/site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php
@@ -28,7 +28,9 @@ class ProjectController extends Controller
      * Render the "New Project" page
      *
      * @param Request $request
+     *
      * @return Response
+     *
      * @throws \Exception
      */
     public function newAction(Request $request)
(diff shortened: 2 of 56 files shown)


site/app/AppKernel.php
```diff
@@ -27,6 +27,7 @@ public function registerBundles()
             new Exercise\HTMLPurifierBundle\ExerciseHTMLPurifierBundle(),
             new Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle(),
             new Sentry\SentryBundle\SentryBundle(),
+            new Http\HttplugBundle\HttplugBundle(),
```
agathver: Where do we use this?

site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php
```diff
@@ -44,12 +46,12 @@ public function newAction(Request $request)
 
         // construct choices for the owner
         $username = $this->getUser()->getUsername();
-        $parentChoices = array($username => 'u_' . $username);
+        $parentChoices = array($username => 'u_'.$username);
         foreach ($this->getUser()->getOrganizationMemberships() as $organizationMembership) {
             if ($organizationMembership->getPermission() === OrganizationMember::PERMISSION_MEMBER or
```
agathver: There appears to be a difference between `and`,`or` and `&&',`||`. Let's standardize on what to use.

http://php.net/manual/en/language.operators.logical.php

site/app/AppKernel.php
imphil: It's a dependency of the updated HWIOAuthBundle

site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php
imphil: Good catch. `||` is the more common one in my world as it has the same precedence as in C (especially regarding the ternary operator). I didn't find a phpcs rule for that and I don't think it's worth writing one, so I've just updated the docs and the code.

agathver:
@imphil, This looks fine to me. If there aren't any more additions, we should go ahead and merge it.

imphil:
Thanks @agathver!

Issue #215, opened 2018-03-06
Make a dashboard page to render after login of a user


sandipbhuyan:
In the present version of librecores after login the user is being redirected to the home page of the application.
- I purpose a dashboard page should be there that will contain the basic information of a user(i.e no of 
 projects, current activity of user in past days, no of user viewed user's profile etc).

agathver:
We don't track anything on LibreCores now, so you will need to implement the tracking too :wink:

PR #250, opened 2018-07-28
Implement dashboard functionality in LibreCores.org

sandipbhuyan:
Resolves issue #215 
This PR adds A dashboard in librecores.org which will be viewed by users after he logs in. This is a platform where we can add different features like user tracking, user interaction etc.

### DIFF
--- a/site/app/Resources/views/base.html.twig
+++ b/site/app/Resources/views/base.html.twig
@@ -87,8 +87,12 @@
               </li>
               <li role="separator" class="divider"></li>
               <li>
-                <a href="{{ path('librecores_project_repo_user_org_view',
-                                 {'userOrOrganization': app.user.username}) }}">
+                <a href="{{ path('librecores.user.dashboard') }}">
+                  Dashboard
+                </a>
+              </li>
+              <li>
+                <a href="{{ path('librecores.user.settings.profile') }}">
                   Your Profile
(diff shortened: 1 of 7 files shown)


imphil:
Please elaborate a bit on the design, I'm confused by the number of pages that show user/org data and settings.

- librecores.org/USERORGNAME is the "public" page of a user or organization. The link to this page has been removed from the menu in your PR.
- librecores.org/user/dashboard is the "dashboard", i.e. a private start page of the user.
- librecores.org/user/settings/profile is the settings page for the user
- librecores.org/USERORGNAME/settings is an alias to the settings page for all users.
- librecores.org/ORGNAME/settings are the settings for an organization.

Inconsistencies:
- For organizations the settings page has a horizontal (on the left) navigation linking to the "view" version (http://www.librecores.devel/openrisc/settings).
- The user settings page has a tab at top saying "settings", but nothing else.
- The public and the private user page do not link to each other.

Can you please:
- Propose a consistent URL scheme.
- Choose navigation either on the top or on the left and use that consistently for user and org pages.
- For a user: link public profile page, dashboard (=private profile page), and settings page with each other using the navigation you chose.
- For a group: link public profile page and settings page.
- Hide the settings navigation if the currently logged-in user does not have appropriate permissions.

PR #257, opened 2018-08-01
Resolve search for classification bug

### LINKED ISSUE #255: Algolia: Search for classification from search results includes HTML
STR:
- Search for a project that has classifications assigned to it.
- The search results show the classifications as part of the result page.
  grafik
- Click on the classification to trigger a new search

Expected result:
- Search results for this classification

Actual result: 
Search string is a HTML string, instead of a plaintext string of the classification. That does not yield any search result.

grafik

sandipbhuyan:
Resolves #255


imphil:
Thanks, looks good!

PR #299, opened 2019-01-18
[CLEANUP] Nuke WordPress

agathver:
It's old, outdated, insecure and we are not going to use it.


agathver:
Done, please check

imphil:
Looks mostly good, but reverts the changes in #298. Can you rebase on top of master to fix that?

imphil:
Thanks @agathver, cleanup work is always appreciated. (And this one makes the repo easier to understand as well with less dead code around.)

Issue #276, opened 2018-08-26
Custom ranking configuration for projects in algolia


sandipbhuyan:
Currently we are not configuring custom ranking in algolia, But we can use the quality matrix stars to rank the projects. According to that the search result will be displayed.
You can refer [Custom ranking algolia](https://www.algolia.com/doc/guides/ranking/custom-ranking/) to set projects ranking.

aquibbaig:
Can I work on this issue?

agathver:
It actually makes sense to perform rank calculation after project update is completed in one of the `RepoCrawler`s. You can extract the logic into a `Service` and call it after `updateRepository()` is run, and update the project entity with the score.

agathver:
Algolia should be updated whenever we do a flush()

sandipbhuyan:
@agathver Can we find any other way to set the ranking. Cause if we are pushing the data to algolia when the project is being inserted into the database then we cannot add custom ranking to the existing projects. I think we can make a ranking calculation from the attributes which can be accessed by the project entity.

agathver:
We can always retrigger a crawling for all projects from the backend side.
A command to calculate metrics for all projects should do it.

aquibbaig:
Ok. @agathver

PR #302, opened 2019-02-09
Custom ranking configuration for projects in algolia

aquibbaig:
This is an initial PR for setting the custom rankings for the projects in Algolia. Please review it and suggest me changes. The metrics are set while creation of a new project. This is for the issue #276


site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php
```diff
@@ -221,6 +221,20 @@ public function viewAction(Request $request, $parentName, $projectName)
             'isHostedOnGithub' => GithubRepoCrawler::isGithubRepoUrl($p->getSourceRepo()->getUrl()),
         ];
 
+        //Retrieve the project and set the code quality metrics
```
sandipbhuyan: Add space after the starting of the comment section `// Retrieve`

site/src/Librecores/ProjectRepoBundle/Entity/Project.php
```diff
@@ -1097,4 +1105,23 @@ public function getClassifications()
...
+     *
+     * @param float $projectMetrics
+     */
+    public function setProjectMetrics($projectMetrics)
+    {
+        $this->projectMetrics = $projectMetrics;
+    }
+
+    /**
+     * Get the code quality metrics for the project
```
sandipbhuyan: New line after description

aquibbaig:
@sandipbhuyan resolved :+1:

site/src/Librecores/ProjectRepoBundle/Entity/Project.php
```diff
@@ -312,6 +312,14 @@ class Project
      */
     private $releases;
 
+    /**
+     * Code quality metrics for the project
+     * @var float
+     *
+     * @ORM\Column(type="float", nullable=true)
```
imphil: What's the motivation for storing this as float? Isn't a (possibly scaled, e.g. multiplied by 10) integer sufficient?

site/src/Librecores/ProjectRepoBundle/Entity/Project.php
```diff
@@ -312,6 +312,14 @@ class Project
...
+     */
+    private $projectMetrics;
```
imphil: The name `projectMetrics` indicates multiple items (plural). That's a simple number you want to store here? How about `score`? (the `project` prefix is already in the name of the table)

site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php
```diff
@@ -221,6 +221,20 @@ public function viewAction(Request $request, $parentName, $projectName)
             'isHostedOnGithub' => GithubRepoCrawler::isGithubRepoUrl($p->getSourceRepo()->getUrl()),
         ];
 
+        // Retrieve the project and set the code quality metrics
```
imphil: I didn't really follow the full discussion, but why is there a database update in the view controller? That looks wrong.

aquibbaig:
@imphil The data which is pushed into algolia is normalised only when we make a function in the entity. However it is not advisable to use services inside the entity itself since the project is already not created. So we cannot access the CommitRepository file inside the Project entity, which is the requirement for the project metrics to be calculated.
And the custom ranking stars are subjected to change as per the changes in the code. This data is updated when we display the project itself by fetching the metadata. So, I hoped that if we run a database migration and set the metrics as and when it is fetched, we can always update the database when they are updated further too. And we custom rank it in algolia accordingly as we call the getter in the entity itself.

Anyways, we can also have a way around if you can suggest one :)

site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php
agathver: My suggestion is to move it to GitRepoCrawler, we can persist it there itself.

agathver:
@aquibbaig We can always add manually re-trigger a crawl of all registered projects. No need to fire a crawl on initial read. We anyway want to do a periodic crawl through a cronjob.

aquibbaig:
@agathver @imphil I tried using the project metrics provider service but the service uses instances of commitRepository and contributorRepository, hence I had to move the logic totally to GitRepoCrawler, where the data for the project metrics is crawled. Locally setting up the environment for me also crawls the scores for other projects too. This PR is for #276 . Please review it.

site/src/Librecores/ProjectRepoBundle/Entity/Project.php
aquibbaig: @imphil The function getCodeQualityScore that calculates the project quality metrics returns a float value according to its definition

imphil: Floats are really tricky to get right in comparisons (and hence sorting), integers are much easier to handle. How about multiplying the metrics value by 100 and storing that as integer value, effectively limiting the precision of the float to two digits.

site/src/Librecores/ProjectRepoBundle/Entity/Project.php
```diff
@@ -312,6 +312,14 @@ class Project
      */
     private $releases;
 
+    /**
+     * Code quality metrics for the project
+     * @var float
+     *
+     * @ORM\Column(type="float", nullable=true)
+     */
+    private $score;
```
imphil: could you change that attribute to `qualityScore` to make it a bit more descriptive?

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -175,11 +179,380 @@ public function updateProject()
...
+        $projectMetrics = $this->getCodeQualityScore($project);
+
+        $project->setScore($projectMetrics);
+
         $this->manager->persist($project);
 
         return true;
     }
 
+    /**
```
imphil: All of this code is copied from `ProjectMetricsProvider.php`. Why don't you inject that class as service into the GitRepoCrawler and use it here?

aquibbaig: @imphil It took some time to figure out inheritance rules among the classes. I guess this fixes #276

aquibbaig:
@imphil I overlooked that float typo. Is this okay now? :D

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -88,6 +98,14 @@ class GitRepoCrawler extends RepoCrawler
 
     private $repoClonePath = null;
 
+    protected $projectMetricsProvider;
+
+    public function __construct(SourceRepo $repo, MarkupToHtmlConverter $markupConverter, ProcessCreator $processCreator, ObjectManager $manager, LoggerInterface $logger, ProjectMetricsProvider $projectMetricsProvider)
```
imphil: - Style: Please split this line into two (at least).
- What makes the project metrics git-specific to be included only in the GitRepoCrawler, and not in the more general RepoCrawler? It looks to me that it should go there.

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -2,14 +2,24 @@
...
+use Librecores\ProjectRepoBundle\Util\MarkupToHtmlConverter;
+use Librecores\ProjectRepoBundle\Util\ProcessCreator;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\Process\Process;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Filesystem\Exception\IOExceptionInterface;
+use Librecores\ProjectRepoBundle\Entity\Project;
+use Librecores\ProjectRepoBundle\Util\StatsUtil;
+use Librecores\ProjectRepoBundle\Util\Dates;
+use Librecores\ProjectRepoBundle\Doctrine\ProjectMetricsProvider;
```
imphil: I don't see code that makes use of all those added classes. Why are they necessary?

site/src/Librecores/ProjectRepoBundle/RepoCrawler/RepoCrawler.php
```diff
@@ -77,6 +78,7 @@ public function __construct(
      * Update the project associated with the crawled repository with
      * information extracted from the repo
      *
+     * @param ProjectMetricsProvider $projectMetricsProvider
```
imphil: that parameter doesn't exist any more.

site/src/Librecores/ProjectRepoBundle/RepoCrawler/RepoCrawler.php
```diff
@@ -3,6 +3,7 @@
 namespace Librecores\ProjectRepoBundle\RepoCrawler;
 
 use Doctrine\Common\Persistence\ObjectManager;
+use Librecores\ProjectRepoBundle\Doctrine\ProjectMetricsProvider;
```
imphil: That's unnecessary as well it seems.

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -88,6 +92,17 @@ class GitRepoCrawler extends RepoCrawler
 
     private $repoClonePath = null;
 
+    public function __construct(
+        SourceRepo $repo,
+        MarkupToHtmlConverter $markupConverter,
+        ProcessCreator $processCreator,
+        ObjectManager $manager,
+        LoggerInterface $logger,
+        $projectMetricsProvider
```
imphil: Is there a reason for this parameter not having a type annotation?

site/src/Librecores/ProjectRepoBundle/RepoCrawler/RepoCrawler.php
```diff
@@ -89,4 +98,14 @@ public function updateSourceRepo()
...
+
+    /**
+     * Access ProjectMetricsProvider class in GitRepoCrawler and GithubRepoCrawler
+     *
+     * @return ProjectMetricsProvider
+     */
+    public function getProjectMetricsProvider()
+    {
+        return $this->projectMetricsProvider;
+    }
```
imphil: Why is that function needed?  Within the subclasses you can access the metrics provider through `$this->projectMetricsProvider` directly.

site/app/DoctrineMigrations/Version20190209100556.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+final class Version20190209100556 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE Project ADD qualityScore INT DEFAULT NULL');
+        $this->addSql('DROP INDEX classification_idx ON ProjectClassification');
```
imphil: I wonder what change caused this index to be dropped?

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
```diff
@@ -2,14 +2,18 @@
...
 use Librecores\ProjectRepoBundle\Entity\ProjectRelease;
+use Librecores\ProjectRepoBundle\Entity\SourceRepo;
 use Librecores\ProjectRepoBundle\Util\FileUtil;
-use Symfony\Component\Process\Process;
+use Librecores\ProjectRepoBundle\Util\MarkupToHtmlConverter;
+use Librecores\ProjectRepoBundle\Util\ProcessCreator;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\Filesystem\Filesystem;
-use Symfony\Component\Filesystem\Exception\IOExceptionInterface;
+use Symfony\Component\Process\Process;
```
imphil: Why all those changes? Were they wrong before? I don't see changes in the code that would make those necessary.

aquibbaig:
Thanks, I'll update it today

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
aquibbaig: Some classes are necessary because I have created a constructor to inject projectmetrics along with all other classes

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
aquibbaig: Both of the approaches will work because projectmetrics are initialized from the RepoCrawler itself thru the constructor. So, I removed the effort of requiring the ProjectMetricsProvider class always and injecting it.

site/src/Librecores/ProjectRepoBundle/RepoCrawler/GitRepoCrawler.php
aquibbaig: For the constructor

imphil:
Thanks @aquibbaig and congratulations to your first contribution! This is now merged and will be deployed with the next push to production.

aquibbaig:
Thanks @imphil

PR #315, opened 2019-04-08
Fix count comparison in UserOrgNameValidator

### LINKED ISSUE #314: Cannot create an user
Creating a user always fails with the message: "The user/organization name is already taken."

Possible cause: Error in `UserOrgNameValidator`

agathver:
Changes introduced in b9f93b5 broke the non-strict comparison used `UserOrgNameValidator`.

Use EntityRepository::count for better abstraction of logic and to avoid a non-strict comparison in our code.

Fix #314

### DIFF
--- a/site/src/Librecores/ProjectRepoBundle/Validator/Constraints/UserOrgNameValidator.php
+++ b/site/src/Librecores/ProjectRepoBundle/Validator/Constraints/UserOrgNameValidator.php
@@ -149,11 +149,7 @@ class UserOrgNameValidator extends ConstraintValidator
 
         // Check the org name against existing usernames
         if ($type === "org") {
-            $q = 'SELECT COUNT(u.id) FROM LibrecoresProjectRepoBundle:User u '.
-                'WHERE u.usernameCanonical = :name';
-            $cntUser = $em->createQuery($q)
-                ->setParameter('name', $name)
-                ->getSingleScalarResult();
+            $cntUser = $em->getRepository('LibrecoresProjectRepoBundle:User')->count([ 'usernameCanonical' => $name ]);
             if ($cntUser !== 0) {
                 return true;
             }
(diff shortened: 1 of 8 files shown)


site/src/Librecores/ProjectRepoBundle/Validator/Constraints/UserOrgNameValidator.php
```diff
@@ -149,24 +149,15 @@ private function userOrOrgNameExists($name, $type)
...
 
         // Check the username against existing org names
         if ($type === "user") {
-            $q = 'SELECT COUNT(o.id) '.
-                'FROM LibrecoresProjectRepoBundle:Organization o '.
-                'WHERE LOWER(o.name) = :name';
-            $cntOrg = $em->createQuery($q)
-                ->setParameter('name', $name)
-                ->getSingleScalarResult();
+            $cntOrg = $em->getRepository('LibrecoresProjectRepoBundle:Organization')->count([ 'name' => $name ]);
```
imphil: You're missing the lowercasing of the org name here compraed to the original code?

agathver: The column name is set to utf8mb4_unicode_ci which performs case insensitive comparisons.

However, I agree on changing it to lower case to be explicit.

agathver: However, the values are always lowercased before passed into this function. `usernameCanonical` and `name` is always lowercased.

imphil: Given that this function is the validator we must ensure that it is as robust as possible. Since we cannot assert on "must be lowercased", we should perform the conversion here as well to make it absolutely certain that we give the right response no matter what gets passed into this function

agathver: Yes, we do that in line 147, we convert all names to lowercase. https://github.com/librecores/librecores-web/pull/315/files/46fb6e06feff990523270acd08f893ab40bbab1a#diff-3124a3d0ca614d0ede52a5306e9bfe1fR147

imphil: But we don't force the name in the database to be lowercased. That's only covered by by the case-insensitive collation, but the `name` field itself can be mixed case. Since we to be case-preserving for the Name field (just like GitHub is) we cannot force the name to be lowercase in the DB, hence we need to lowercase during the comparison.

imphil:
@agathver what's blocking this?

agathver:
@imphil This is good to go.

Earlier I was stuck with the functional tests I have been writing

PR #319, opened 2019-04-12
Skip instead of marking non-existent project IDs as failed

### LINKED ISSUE #317: RabbitMQ daemon crashes if a queued project is deleted before it has been processed
Steps to reproduce:

1. Add a project
2. Remove the project (from the database).
3. Daemon crashes with a null pointer exception

Though we do not have a delete project UI as of now, we can simulate it by truncating the tables.

agathver:
Fixes #317


imphil:
LGTM

PR #320, opened 2019-04-12
Fix classification tree generation

### LINKED ISSUE #318: Opening project settings fails
Fails with a null pointer exception at [ProjectController.php#L332](https://github.com/librecores/librecores-web/blob/master/site/src/Librecores/ProjectRepoBundle/Controller/ProjectController.php#L332)

Possible reasons:
* Compare and assign logic is wrong

agathver:
Avoids Null Pointer Exception, which prevents classification tree from generating.

Fixes #318


imphil:
LGTM, thanks!

PR #326, opened 2019-04-18
Reorganize tests into unit tests and functional tests

agathver:


agathver:
This is still WIP

agathver:
Attempting functional tests in PHP was a bad idea.

Issue #327, opened 2019-04-22
Test Automation


agathver:
Meta issue to track testing progress

- [ ] Unit tests
	- [ ] Tests for utilities
	- [ ] Tests for services
	- [ ] Tests for symfony extensions (forms, validators etc.)
 	- [ ] Tests for controllrs
- [ ] Integration tests
	- [ ] Controller test with simulated HTTP requests
- [ ] Functional tests
	- [ ] Basic tests
		- [ ] Home page renders
		- [ ] Static pages render
		- [ ] Browse projects when logged out
	- [ ] User tests
		- [ ] Login through email
		- [ ] Login through Github
		- [ ] Add project
		- [ ] Crawler crawls the project successfully
		- [ ] Project metrics are displayed
	- [ ] Extended user tests
		- [ ] Register
		- [ ] Password Reset
		- [ ] Login through Google
		- [ ] Change password
		- [ ] Change name
		- [ ] Browse projects
	- [ ] Extended project tests
		- [ ] Add GitHub project
		- [ ] Algolia index is updated succesfully
		- [ ] Change project settings
		- [ ] Add organization
		- [ ] Add project to organization

PR #328, opened 2019-04-23
Selenium Tests

agathver:
First of the automation tests to come.

Add functional tests in a real browser using something like Selenium. Initially should cover basic functionality only

Refer #327


imphil:
Looks good, thanks!

PR #334, opened 2019-04-25
Ansible: Install frontend JS dependencies

imphil:
Install yarn and nodejs as prerequisite to switch from assetic to
webpack encore and other frontend technologies.

@aquibbaig that should be enough to get you started with webpack encore; let me know if you need something else installed.

### DIFF
--- a/ansible/roles/web/tasks/frontend-js.yml
+++ b/ansible/roles/web/tasks/frontend-js.yml
@@ -1,0 +1,27 @@
+- name: Add yarn repository key
+  apt_key:
+    url: https://dl.yarnpkg.com/debian/pubkey.gpg
+    state: present
+
+- name: Add yarn repository
+  apt_repository:
+    repo: deb https://dl.yarnpkg.com/debian/ stable main
+    state: present
+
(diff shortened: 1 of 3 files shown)


aquibbaig:
Thank you @imphil !

ansible/roles/web/tasks/frontend-js.yml
```diff
@@ -0,0 +1,27 @@
...
+    state: present
+
+- name: Add NodeSource repository key
+  apt_key:
+    url: https://deb.nodesource.com/gpgkey/nodesource.gpg.key
+    state: present
+
+- name: Add node.js repository from NodeSource
+  apt_repository:
+    repo: deb https://deb.nodesource.com/node_12.x xenial main
```
agathver: I would stick to NodeJS 10 till NodeJS 12 is promoted as LTS. Node 12 was just released a few days ago and it still may have regressions with the dependencies with WebPack.

Module authors usually take a while to ensure compatibility with a new release.

imphil:
Thanks for the feedback @agathver. I've updated the PR to use Node.js v10

PR #341, opened 2019-04-28
Fix flaky tests in UserOrgValidator

### LINKED ISSUE #342: UserOrgValidatorTest is flaky
Test breaks randomly due to two validation errors being raised which is unexpected at that context.

agathver:
UserOrgValidatorTest uses property testing and it used to fail
when a string of single character was randomly generated. However,
single character strings were to be caught by a different rule.

Hence, this change fixes minimum length of random string to 2

Fixes #342


imphil:
Thanks!

Issue #292, opened 2019-01-03
Switch away from assetic


imphil:
The symfony assetic bundle is being deprecated, the suggested replacement is Webpack Encore (https://symfony.com/doc/current/frontend.html). We should consider switching as well to give us (in addition to a maintained library) a more "standard" frontend experience and enable the use of frameworks like vue.js for some frontend pieces.

aquibbaig:
If we implement this, it will solve #260 as well.

aquibbaig:
#353 currently is in progress to migrate from Assetic to webpack. 
There are a few things we need to tackle for configuring page-specific CSS files. Each file that webpack builds is an independent file of its own with all its dependencies intact. We have some twig files in which we need to pass twig filters to the jquery functions. 

**For example:**
> $('#{{ form.descriptionText.vars.id }}').trumbowyg(trumbowyg_config);
  $('#{{ form.licenseText.vars.id }}').trumbowyg(trumbowyg_config);

The above code here is from `LibrecoresProjectRepoBundle:Project:settings.html.twig` which uses a twig parameter `{{ form.descriptionText.vars.id }}`. For bundling it with webpack, this parameter needs to be accessed inside a single javascript file. I believe refactoring all the javascript in this page, i.e. `LibrecoresProjectRepoBundle:Project:settings.html.twig` to a single js file can be an example to refactor other similar situations. Can anyone help me out in this? @agathver @imphil @sandipbhuyan

agathver:
@imphil Do we even use trumbowyg anywhere?

imphil:
> page-specific CSS files.

did you mean JS files?

imphil:
> @imphil Do we even use trumbowyg anywhere?

yes, in the project settings page for the LICENSE and README text boxes

aquibbaig:
@agathver in the LICENSE and DESCRIPTION fields

agathver:
@aquibbaig Here's a suggested approach:
- Assign a class (let's say: `.trumbowyg-textbox`) to each of the text areas that use trumbowyg.
- Use a jQuery selector to find any `.trumbowyg-textbox` in a page.
- Use [dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Dynamic_imports) to load trumbowyg and initialize them.

```js
let textareas = $('.trumbowyg-textbox);
if(textareas.length) {
    import('/modules/my-module.js')
        .then(() => {
[...]

PR #344, opened 2019-04-30
Webpack

aquibbaig:
Solves #292. This is an initial PR regarding implementation of webpack. Will proceed after reviews regarding the approach.
> TODO: 
- Remove Assetic completely
- Bundle js files using `encore`


site/app/Resources/views/base.html.twig
```diff
@@ -11,17 +11,12 @@
...
-        "assets/css/librecores.css"
-    %}
-        <link rel="stylesheet" href="{{ asset_url }}" />
-    {% endstylesheets %}
-  {% endblock %}
+    {# 'app' must match the first argument to addEntry() in webpack.config.js #}
+    {{ encore_entry_link_tags('app') }}
 
+    <!-- Renders a link tag (if your module requires any CSS)
+         <link rel="stylesheet" href="/build/app.css"> -->
```
imphil: Is that needed for CSS to get included?

site/app/Resources/views/base.html.twig
```diff
@@ -155,6 +150,12 @@
       <script src="{{ asset_url }}"></script>
     {% endjavascripts %}
 
+    {{ encore_entry_script_tags('app') }}
+
+    <!-- Renders app.js & a webpack runtime.js file -->
```
imphil: Does this really render the files, or just include them?

site/app/config/webpack_encore.yml
```diff
@@ -0,0 +1,20 @@
...
+    # if you have multiple builds:
+    # builds:
+    # pass "frontend" as the 3rg arg to the Twig functions
+    # {{ encore_entry_script_tags('entry1', null, 'frontend') }}
+
+  # frontend: '%kernel.project_dir%/public/frontend/build'
+
+  # Cache the entrypoints.json (rebuild Symfony's cache when entrypoints.json changes)
+  # Available in version 1.2
+  #cache: '%kernel.debug%'
```
imphil: nit: Trailing newline missing.

site/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "devDependencies": {
+    "@symfony/webpack-encore": "^0.27.0"
+  },
+  "dependencies": {
+    "node-sass": "^4.12.0",
+    "sass-loader": "^7.1.0"
+  }
+}
```
imphil: Should there be a package lock file as well?

site/web/assets/scss/bootstrap-librecores.scss
```diff
@@ -14,4 +14,4 @@ $screen-md-min:              $screen-md;
 //** Point at which the navbar becomes uncollapsed.
 $grid-float-breakpoint:     $screen-md-min;// Import Bootstrap for Sass
 
-@import "bootstrap";
+@import "bootstrap";
```
imphil: nit: Trailing newline missing.

site/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+    // enables hashed filenames (e.g. app.abc123.css)
+    .enableVersioning(Encore.isProduction())
+
+// uncomment if you use TypeScript
+//.enableTypeScriptLoader()
+
+// uncomment if you use Sass/SCSS files
+  .enableSassLoader()
+
+// uncomment if you're having problems with a jQuery plugin
```
imphil: Do we have problems with jQuery? If so, what problems do we have?

site/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+
+    .cleanupOutputBeforeBuild()
+    .enableSourceMaps(!Encore.isProduction())
+    // enables hashed filenames (e.g. app.abc123.css)
+    .enableVersioning(Encore.isProduction())
+
+// uncomment if you use TypeScript
+//.enableTypeScriptLoader()
+
+// uncomment if you use Sass/SCSS files
```
imphil: You can remove this comment.

site/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+    //.addEntry('page2', './assets/js/page2.js')
+
+    // will require an extra script tag for runtime.js
+    // but, you probably want this, unless you're building a single-page app
+    .enableSingleRuntimeChunk()
+
+    .cleanupOutputBeforeBuild()
+    .enableSourceMaps(!Encore.isProduction())
+    // enables hashed filenames (e.g. app.abc123.css)
+    .enableVersioning(Encore.isProduction())
```
imphil: You enable versioned files here (which is good for production), but always include `app.js` above. That probably won't work?

imphil:
I had an initial look at the code only (didn't run it so far) and added a couple comments. Could you have a look at it?

site/webpack.config.js
agathver: I think we may have. Most of the jQuery plugins we use (Select2, Algolia) are configured to use jQuery as a global than require it.

site/app/Resources/views/base.html.twig
```diff
@@ -155,6 +150,12 @@
...
+    <script src="/build/runtime.js"></script>
```
agathver: If you are using `{[ encore_entry_script_tags('...') }}` then you do not need explicit `script` tags. The macro will render it for you.

site/web/assets/scss/bootstrap-librecores.scss
```diff
@@ -3,7 +3,7 @@ $link-hover-color: 		#173e43;
 $font-family-sans-serif:  'Open Sans', sans-serif;
 $font-family-monospace:   "Source Code Pro", Menlo, monospace;
 $font-size-base:          15px;
-$icon-font-path: "../fonts/";
+$icon-font-path: "../../fonts/";
```
agathver: Alignment issue here?

site/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+    /*
+     * ENTRY CONFIG
+     *
+     * Add 1 entry for each "page" of your app
+     * (including one that's included on every page - e.g. "app")
+     *
+     * Each entry will result in one JavaScript file (e.g. app.js)
+     * and one CSS file (e.g. app.css) if you JavaScript imports CSS.
+     */
+    .addEntry('app', ['./web/assets/scss/librecores.scss',
```
agathver: Instead of adding all files here, create an `app.scss` and include CSS files inside it.

site/webpack.config.js
aquibbaig: @imphil I didn't get what you mean to say here.

site/app/Resources/views/base.html.twig
aquibbaig: I don't think so. We have included all of them in the `app` argument

site/package.json
aquibbaig: Why do we need a `package lock`. We have a `yarn.lock`.

site/webpack.config.js
aquibbaig: That's a good idea. Thanks!

site/app/Resources/views/base.html.twig
aquibbaig: And it is automatically included from there. I checked `dev tools
`

site/app/Resources/views/base.html.twig
aquibbaig: Includes the `build` files

site/webpack.config.js
agathver: @aquibbaig You have a fixed link to "build/app.js" which will not work. Use `encore_entry_script_tags`

site/app/Resources/views/base.html.twig
```diff
@@ -11,17 +11,10 @@
...
-        <link rel="stylesheet" href="{{ asset_url }}" />
-    {% endstylesheets %}
-  {% endblock %}
+    {# 'app' must match the first argument to addEntry() in webpack.config.js #}
+    {{ encore_entry_link_tags('app') }}
 
+  {% endblock %}
   <!-- HTML5 shim and Respond.js IE8 support of HTML5 elements and media queries -->
   <!--[if lt IE 9]>
     {% javascripts filter="?jsqueeze" output="js/app_head.js" '@js_head'%}
```
imphil: Are those JS files still needed?

site/app/Resources/views/base.html.twig
```diff
@@ -155,6 +148,8 @@
       <script src="{{ asset_url }}"></script>
     {% endjavascripts %}
 
+    {{ encore_entry_script_tags('app') }}
```
imphil: What happens to the JS files to be included at the end of the file? (i.e. most of them?)

site/webpack.config.js
imphil: So please mention this fact in the comment.

site/app/Resources/views/base.html.twig
aquibbaig: I think we add js files at the end because we don't have to worry if the DOM has been loaded or not, everything will work fine.

imphil: We add JS files at the end because it's good for performance. We only add JS files to the header which have need to be there. This current setup should be preserved after we move to webpack.

imphil:
I gave this a quick try in a fresh VM (`vagrant destroy; vagrant up`) and it showed me "An exception has been thrown during the rendering of a template ("Could not find the entrypoints file from Webpack: the file "/var/www/lc/site/web/build/entrypoints.json" does not exist.")."

How is `entrypoints.json` created?

aquibbaig:
I am closing this Pull Request and reopening through another branch because this is not the correct way of configuring encore. I am working remotely in another branch

imphil:
You do not need to close a PR, you can (and should) update it by pushing to the branch as needed, including force pushing when you do a rebase.

aquibbaig:
Okay. Done!

agathver:
> I gave this a quick try in a fresh VM (`vagrant destroy; vagrant up`) and it showed me "An exception has been thrown during the rendering of a template ("Could not find the entrypoints file from Webpack: the file "/var/www/lc/site/web/build/entrypoints.json" does not exist.")."
> 
> How is `entrypoints.json` created?

It is created when we either run `yarn build` or `yarn dev-server`. You can run this either inside the VM or outside. Outside the VM will be much much faster & more performant as it will be able to use `inotify/fsevents` which is not available in NFS shares.

aquibbaig:
@agathver @imphil Had closed this PR. We can move to #353 for discussions

PR #356, opened 2019-05-10
Inform user about missing GitHub access when adding projects

### LINKED ISSUE #305: Re-request GitHub credentials if we get a "bad credentials" exception
I get a _500 Internal Server Error_ when I try to add a new project by clicking my username in the right top corner.

agathver:
Shows a button to reconnect to GitHub

Fixes #305

### DIFF
--- a/site/app/config/config.yml
+++ b/site/app/config/config.yml
@@ -34,7 +34,7 @@ framework:
     cache:
         pools:
             # Cache pool for all GitHub API requests done through
-            # GithubApiService
+            # GitHubApiService
             app.cache.githubapi:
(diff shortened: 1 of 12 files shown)


imphil:
So by this PR, if we detect an auth failure when talking with GitHub, we treat this as temporary error and do not disable the GitHub linkage permanently. Sounds good as a first stab at this problem, but we most likely need to go beyond that in a future change (e.g. disconnecting from GitHub permanently if errors accumulate, or if we have clear indication that this isn't a temporary problem but the user permanently removed our GitHub connection)

PR #364, opened 2019-05-26
Separate GitHub crawler

### LINKED ISSUE #346: Gather github metrics independent of crawling
For projects hosted in GitHub we support additional metadata like stars, issues and pull requests. 

In the current setup, these are refreshed only when a repo crawl is performed which is during a new commit to the project. Thus, most of our GitHub data is stale. 

By decopuling the crawlers we can update github data more frequently.

agathver:
Separates GitHub metadata into its own crawler.
Entirely reworks the way the crawlers work.

Implementation details:
* GitHub metadata is treated as optional data and hence we don't show "project is in processing" if the project is in the Github metadata queue, but has completed processing of normal metadata (this is rare)

Fixes #346


site/src/Controller/ProjectController.php
```diff
@@ -273,7 +275,7 @@ public function viewAction(
             'contributorsGraph' => $this->makeGraph(
                 $projectMetricsProvider->getContributorHistogram($p, Dates::INTERVAL_YEAR)
             ),
-            'isHostedOnGithub' => GithubRepoCrawler::isGithubRepoUrl($p->getSourceRepo()->getUrl()),
+            'isHostedOnGithub' => preg_match_all(self::GH_REGEX,$p->getSourceRepo()->getUrl()),
```
imphil: Can you move the regex into the SourceRepo and/or GitSourceRepo class, e.g. as `getHostingSite()`? We don't want to have this check in multiple places.

site/src/Consumer/UpdateGitHubMetadataConsumer.php
```diff
@@ -0,0 +1,247 @@
...
+            // next requests.
+            $this->logger->error(
+                "Processing of repository resulted in an ".get_class($e)
+            );
+            $this->logger->error('Message: '.$e->getMessage());
+            $this->logger->error('Trace: '.$e->getTraceAsString());
+
+            return false;
+        }
+    }
```
imphil: Most code in this consumer is copied from UpdateProjectInformation. Can you make a common base class for these two, e.g. `ProjectConsumer` that contains the shared code in `execute()`, and calls a `executeProjectUpdate(Project)` function or something similar?

site/src/Consumer/UpdateGitHubMetadataConsumer.php
```diff
@@ -0,0 +1,247 @@
...
+     */
+    private function updateGithubMetadata(Project $project)
+    {
+        try {
+            $this->logger->info("Enriching project: {$project->getFqname()} with GitHub metadata");
+            $client = $this->getGithubClient($project);
+
+            $repoUrl = $project->getSourceRepo()->getUrl();
+
+            $isGithubRepo = preg_match(static::GH_REGEX, $repoUrl, $matches);
```
imphil: see below: please put this check into the SourceRepo class or its subclasses.

agathver:
@imphil Rebased and updated

site/src/Consumer/AbstractProjectUpdateConsumer.php
```diff
@@ -0,0 +1,95 @@
...
+                // table
+                $this->logger->error(
+                    "Unable to update project with ID $projectId: project does "
+                    ."not exist."
+                );
+
+                return true; // don't requeue
+            }
+
+            return $this->processProject($project);
```
imphil: You need to return true or false here to signal "requeue" or "not requeue". Please update the docs for `AbstractProjectUpdateConsumer::processProject` and the implementations (which currently return nothing mostly).

PR #366, opened 2019-05-28
Fix ansible deprecations

### LINKED ISSUE #297: Fix ansible deprecations
Ansible has changed the behavior of `apt` and now accepts multiple packages in a single command. The old way of running apt with `with_items` has been deprecated and is scheduled to be removed soon.

We have around 5-6 usages of apt with `with_items` we need to migrate them to the new syntax.

agathver:
Switch to new syntax for `apt` module which now accepts a list of packages under `pkg`.

NOTE: The warning about `use_https` is a false-positive and hence is left as-is

Fixes #297

### DIFF
--- a/ansible/roles/web/tasks/php.yml
+++ b/ansible/roles/web/tasks/php.yml
@@ -19,7 +19,10 @@
 # OS packaged PHP extensions
 - name: Install packaged PHP extensions
   become: true
-  apt:  package=php{{php_version}}-{{ item }} state=present
+  apt:
+    package:
+        - php{{php_version}}-{{ item }}
+    state: present
   with_items: "{{ php_extensions_packaged }}"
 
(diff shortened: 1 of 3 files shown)


ansible/roles/web/tasks/php.yml
```diff
@@ -31,7 +34,10 @@
 
 - name: Install packaged PHP extensions for development
   become: true
-  apt:  package=php{{php_version}}-{{ item }} state=present
+  apt:
+    package:
+        - php{{php_version}}-{{ item }}
+    state: present
```
imphil: That nicely works around the deprecation but doesn't solve the issue of combining all packages into a single `apt` call in the future (That's what the warning is all about). Anyways, let's take it for now, all other options are extremely ugly.

PR #372, opened 2019-06-02
Show error when we fail to install a GitHub webhook

### LINKED ISSUE #371: Add  GItHub project fails with 500 if we cannot install install webhooks
Title says all.

We should handle it gracefully and report an error to the user

agathver:
Closes #371

### DIFF
--- a/site/app/Resources/views/layout.html.twig
+++ b/site/app/Resources/views/layout.html.twig
@@ -22,8 +22,11 @@
   {% if app.request.hasPreviousSession %}
     {% for type, messages in app.session.flashbag.all() %}
       {% for message in messages %}
-        <div class="alert alert-{{ type }}" role="alert">
-          {{ message|raw }}
+        <div class="alert alert-{{ type }} alert-dismissable" role="alert">
+          <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
+          {{ message | trans | raw }}
         </div>
       {% endfor %}
     {% endfor %}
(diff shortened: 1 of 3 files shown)


site/app/Resources/views/layout.html.twig
```diff
@@ -22,8 +22,11 @@
   {% if app.request.hasPreviousSession %}
     {% for type, messages in app.session.flashbag.all() %}
       {% for message in messages %}
-        <div class="alert alert-{{ type }}" role="alert">
-          {{ message|raw }}
+        <div class="alert alert-{{ type }} alert-dismissable" role="alert">
+          <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
+          {{ message | trans | raw }}
```
imphil: We don't have translated strings anywhere, just hardcode it here in English.

site/src/Controller/ProjectController.php
```diff
@@ -162,13 +164,17 @@ public function newAction(
...
-            );
+            try {
+                $this->populateProjectFromForm(
+                    $p,
+                    $form,
+                    $githubApiService,
+                    $organizationRepository,
+                    $userManager
+                );
+            } catch (ErrorException $e) {
```
imphil: Can we find a more focused Exception to catch, or event give the user bits of the error message we got from GitHub?

site/app/Resources/views/layout.html.twig
agathver: I use some translations as the lines were being too long when included in code and were a bit testing-unfriendly

imphil:
Thanks, LGTM!

Issue #374, opened 2019-06-02
[GSoC 2019] Store notifications in the database


aquibbaig:
- As discussed, the database schema needs to be updated for storing notifications of users. This issue will be an immediate follow up of #373. The notifications are supposed to be related to the Users table, so we should keep that in mind. 
- Also, DataFixtures are needed to be written for the above case.

### Tasks involved in this issue
- [x] Extend the database schema
- [x] Store Notifications in database
- [ ] Write DataFixtures for the above

### Further goals
 - This issue should deal with persisting the generated notification to the database so that they can be fetched to the UI afterward.

imphil:
Try using https://packagist.org/packages/mgilet/notification-bundle

aquibbaig:
The bundle above mentioned by @imphil comes with 3 class entities affecting our database schema

1. NotifiableEntity
2. NotifiableNotifications
3. Notification
- `Notifiable` entity is the entity you want notifications for, for example: User entity in our case
- `NotifiableNotification` entity holds the properties of the notifications i.e. if it has been seen. This entity is associated *ManyToOne* with Notifiable entity and Notification itself.
- `Notification` entity is the actual entity that holds the notification and it's fields like message, subject and so on. In our project, the Notification entity has been overriden by `AppNotification` entity as per our requirement to hold new fields such as the type of the Notification.

aquibbaig:
![Entity](https://user-images.githubusercontent.com/26324376/60724352-c81daa00-9f53-11e9-890b-6ff756397875.jpg)

@imphil While creating an entity for Notifications, the following idea came to my mind:
- If we want to notify like many people with the same notification(such as notifying people in the whole org about a project creation), we need to make recipients as Array Collections and map it to the Users. While I have little knowledge about mappings as I am always confused in them, I have drawn out a basic structure, if that's okay. Please let me know what you think

imphil:
It's better to have multiple `Notification`s in this case. It makes things much more complex, e.g. when you add status flags to a notification (like "has been read"). So 1:1 Notification : Recipient.

PR #378, opened 2019-06-05
create and store notifications

### LINKED ISSUE #373: [GSoC 2019] Generate notifications from events
- This is based upon the design as discussed in #363
 - This issue deals with the first implementation goal, which is to generate notifications from events. The basic idea behind this is that for every user action for which a notification can be associated, we must generate a notification entity instance. 
 - The events that we want to support and data fields in the notification and their types are also discussed in #363.

### Constraints 
 - Notifications will be generated only if the User is logged on to the website

### Tasks associated
- [x] Associate actionable events with a suitable notification service to generate notifications

### Further goals
The notifications thus generated will be further stored in the database for the user

aquibbaig:
Fixes #373 and #374. 
Haven't tried it on the RepoCrawler, just was testing things out against one of the simpler mentioned events i.e. Creation of the Project. This is an initial PR which mostly deals with the approach that we want to proceed things with

### DIFF
--- a/site/src/Controller/ProjectController.php
+++ b/site/src/Controller/ProjectController.php
@@ -10,7 +10,6 @@ use App\Entity\Project;
 use App\Entity\ProjectClassification;
 use App\Entity\User;
 use App\Form\Type\ProjectType;
-use App\RepoCrawler\GithubRepoCrawler;
 use App\Repository\OrganizationRepository;
 use App\Repository\ProjectRepository;
 use App\Service\GitHub\AuthenticationRequiredException;
@@ -23,6 +22,7 @@ use DateTimeImmutable;
 use Doctrine\ORM\NonUniqueResultException;
 use Exception;
 use FOS\UserBundle\Model\UserManagerInterface;
+use Github\Exception\ErrorException;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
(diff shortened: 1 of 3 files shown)


site/app/Resources/views/notification/notification_list.html.twig
```diff
@@ -0,0 +1,22 @@
...
+                    <hr/>
+                {% endfor %}
+            {% else %}
+                <div>
+                    No notifications for you yet!
+                </div>
+            {% endif %}
+        </div>
+    </body>
+</html>
```
imphil: Please indent HTML/JS/CSS with two spaces.

site/app/config/routing.yml
```diff
@@ -40,3 +40,7 @@ librecores.user.settings.password:
 librecores_user_resend_confirmation:
     path: /user/resend-confirmation
     controller: App\Controller\UserController::resendConfirmationEmailAction
+
+mgilet_notification:
+    resource: "@MgiletNotificationBundle/Controller/"
+    prefix: /notifications
```
imphil: This path isn't available unless you explicitly exclude it from the UserOrgNameValidator (it could clash with a username/orgname at the same path). What is this routed for? Should it to go `/user/notification`?

site/src/Controller/ProjectController.php
```diff
@@ -174,6 +175,20 @@ public function newAction(
             $em->persist($p);
             $em->flush();
 
+            // project was created, send Notification
+            $manager = $this->get('mgilet.notification');
```
imphil: Inject this service please instead of using `$this->get`.

site/src/Entity/AppNotification.php
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * @ORM\Entity
+ *
+ * @package App\Entity
+ */
+class AppNotification extends NotificationModel implements NotificationInterface
+{
+    /**
+     * Overrides the existing Notifications entity
```
imphil: How does this comment match the `type` attribute?

Issue #379, opened 2019-06-07
[GSoC 2019] Decide a routing mechanism for notifications


aquibbaig:
Based on the recent discussion, we need to set up how to parse a notification **type** and then route it to its specific path(If its email type, send a mail or if it's web type, store in the db and fetch to the frontend). @imphil suggested to implement it using a cronjob or rabbitmq.  

- I guess we won't have that much traffic  and sending a notification won't take that many of processing overhead, so what about this? 
```
1. $notif = $manager->createNotification('Notification subject', 'Some random text', 'https://google.fr/', 'web');  
2. sendNotification($notif);
```
This means after a notification is created -> send it accordingly in the next step. If it's web type store in the database. If it's email type, don't store and just send a mail. 
@agathver @imphil @sandipbhuyan  Let me know your thoughts

imphil:
Questions:
- Where/how do you store the routing configuration? This can be global (all messages of type x to go email), and user-specific (user disabled getting emails about type y).
- You cannot synchronously send emails, you need to queue the email send requests somehow. You didn't really answer that question in issue #363 and that's "haunting" you now.

Look at the picture below for an architecture I'd use. It doesn't answer the routing part in code yet, it only shows you where to hook it up (in the consumer). 

![image](https://user-images.githubusercontent.com/1467123/59163316-43c43c80-8af7-11e9-9cca-d53e5e747f10.png)

You can have a look at the existing crawler code to see RabbitMQ and consumer/producers in action. It makes sense to just re-use this infrastructure and do the notification system in a similar way.


Use this proposed architecture as a starting point to further refine it. Make use of drawing such as this one to convey what you exactly mean, and be specific.

PR #353, opened 2019-05-10
Migrate to webpack 

aquibbaig:
Preliminary commit. Bundled files in the base template

PR #381, opened 2019-06-12
WIP: Notification System

aquibbaig:
This commit is WIP. Currently aimed at fixing #373 #374 #379

### DIFF
--- a/site/.gitignore
+++ b/site/.gitignore
@@ -22,3 +22,4 @@
 /node_modules/*
 /web/build/*
 /yarn-error.log
+/assets/js/frontend-config.json
--- a/site/app/AppKernel.php
+++ b/site/app/AppKernel.php
@@ -24,6 +24,7 @@ class AppKernel extends Kernel
             new Http\HttplugBundle\HttplugBundle(),
             new Algolia\SearchBundle\AlgoliaSearchBundle(),
             new \Symfony\WebpackEncoreBundle\WebpackEncoreBundle(),
+            new Mgilet\NotificationBundle\MgiletNotificationBundle(),
         );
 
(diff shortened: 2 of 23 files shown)


aquibbaig:
Screenshot 2019-06-13 at 3 30 16 PM

Will be coming up with a design for the UI today at the earliest, but this is what is being achievable rn, with all the configuration as discussed @imphil @agathver

aquibbaig:
Screenshot 2019-06-14 at 10 18 54 PM

## Status

- Added Mark as Read and Remove Notifications with their specific handlers. Note that these features do not depend on other parameters than the notification itself such as user, project, etc. So to make it independent, it is wiser to create another Controller with specific routes to handle it.
- Also, the color and styles are subject to change and to be moved to an independent file after #353

site/.gitignore
```diff
@@ -22,3 +22,4 @@
 /node_modules/*
 /web/build/*
 /yarn-error.log
+/assets/js/frontend-config.json
```
imphil: Please do a separate PR for changes like this, they should be merged independent of this PR, which is more experimental.

imphil:
To clarify what I meant with "That forwarding should happen through RabbitMQ, not by calling two handlers from a single consumer." please have a look at this picture:

image

That's the same setup we have for our repo crawler currently.

aquibbaig:
@imphil Found some time out and made the required architectural changes. Probably have left out some PhPDocs comments. will look it up when they allow me to

aquibbaig:
This Pull Request will be open to discussions only

imphil:
This has now been merged through individual PRs, closing.

PR #382, opened 2019-06-26
[GSoC 2019] Implementation of a Notification System for Librecores

aquibbaig:
Aimed at fixing #373 #374 currently. This will be a WIP for now.

### DIFF
--- a/site/app/config/packages/old_sound_rabbit_mq.yml
+++ b/site/app/config/packages/old_sound_rabbit_mq.yml
@@ -19,6 +19,9 @@ old_sound_rabbit_mq:
         update_project_info:
             connection:       default
             exchange_options: {name: 'update-project-info', type: fanout}
+        notification:
+            connection:       default
+            exchange_options: {name: 'notification', type: fanout}
     consumers:
         update_project_info:
             connection:       default
@@ -30,3 +33,14 @@ old_sound_rabbit_mq:
             exchange_options: {name: 'update-project-info', type: fanout}
             queue_options:    {name: 'update-github-metadata'}
             callback:         update_github_metadata_consumer
+    multiple_consumers:
+        notification:
+            connection:       default
+            exchange_options: {name: 'notification', type: fanout}
+            queues:
+                web_notification:
+                    name:     'web-notification'
+                    callback: web_notification_consumer
+                email_notification:
+                    name:     'email-notification'
+                    callback:  email_notification_consumer
--- a/site/app/config/services.yml
+++ b/site/app/config/services.yml
@@ -27,6 +27,8 @@ services:
     # Required legacy-style aliases
     update_project_info_consumer: '@App\Consumer\UpdateProjectInformation'
     update_github_metadata_consumer: '@App\Consumer\UpdateGitHubMetadataConsumer'
+    web_notification_consumer: '@App\Consumer\WebNotificationConsumer'
+    email_notification_consumer: '@App\Consumer\EmailNotificationConsumer'
 
     # Explicit configurations
     # TODO: Investigate and deprecate these in favor of auto-wiring
--- a/site/src/Consumer/AbstractNotificationConsumer.php
+++ b/site/src/Consumer/AbstractNotificationConsumer.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace App\Consumer;
+
+use App\Util\Notification;
+use OldSound\RabbitMqBundle\RabbitMq\ConsumerInterface;
+use PhpAmqpLib\Message\AMQPMessage;
+use Psr\Log\LoggerInterface;
+use Exception;
+
+/**
+ * Class AbstractNotificationConsumer
+ *
+ * Base class for all the Notification Consumers that we have
+ *
+ * @author Aquib Baig <aquibbaig97@gmail.com>
+ */
+abstract class AbstractNotificationConsumer implements ConsumerInterface
+{
+    /**
+     * @var LoggerInterface
+     */
+    protected $logger;
+
+    /**
+     * @var Notification $notification
+     */
+    protected $notification;
+
+    /**
+     * AbstractNotificationConsumer constructor
+     *
+     * @param LoggerInterface $logger
+     */
+    public function __construct(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
+    }
+
+    /**
+     * @param AMQPMessage $msg
+     *
+     * @return bool
+     *
+     * @throws Exception
+     */
+    public function execute(AMQPMessage $msg)
+    {
+        $this->notification = unserialize($msg->body);
+        try {
+            if ($this->shouldHandle()) {
+                return $this->handle();
+            }
+
+            return true; //don't requeue
+        } catch (Exception $e) {
+            // Log out exceptions if they occur, and keep Consumers running for
+            // next requests
+            $this->logger->error(
+                "Processing the Notification resulted in an ".get_class($e)
+            );
+            $this->logger->error("Message: ".$e->getMessage());
+            $this->logger->error("Trace: ".$e->getTraceAsString());
+
+            return false;
+        }
+    }
+
+    /**
+     * Should a Notification be handled by a the Consumer?
+     *
+     * @return bool
+     */
+    protected function shouldHandle()
+    {
+        return true;
+    }
+
+    /**
+     * Actually processes a Notification to its sink
+     *
+     * @return mixed
+     */
+    abstract protected function handle():bool;
+}
+
+?>
--- a/site/src/Consumer/EmailNotificationConsumer.php
+++ b/site/src/Consumer/EmailNotificationConsumer.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace App\Consumer;
+
+use App\Util\Notification;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Sends out Notifications over email
+ *
+ * @author Aquib Baig <aquibbaig97@gmail.com>
+ */
+class EmailNotificationConsumer extends AbstractNotificationConsumer
+{
+    /**
+     * EmailNotificationConsumer constructor
+     *
+     * @param LoggerInterface $logger
+     */
+    public function __construct(LoggerInterface $logger)
+    {
+        parent::__construct($logger);
+    }
+
+    /**
+     * Handles an email Notification
+     *
+     * @return bool
+     */
+    protected function handle(): bool
+    {
+        // TODO: Process the Notification($this->notification)
+        echo "Email Notification Consumer";
+
+        return true;
+    }
+}
+
+?>
--- a/site/src/Util/Notification.php
+++ b/site/src/Util/Notification.php
@@ -1,0 +1,177 @@
+<?php
+
+namespace App\Util;
+
+use App\Entity\User;
+use DateTime;
+
+/**
+ * Data class for all Notifications
+ *
+ * @author Aquib Baig <aquibbaig97@gmail.com>
+ */
+class Notification
+{
+    /**
+     * Subject of the Notification
+     *
+     * @var string $subject
+     */
+    protected $subject;
+
+    /**
+     * Notification message
+     *
+     * @var string $message
+     */
+    protected $message;
+
+    /**
+     * Recipient of the Notification
+     *
+     * @var User $recipient
+     */
+    protected $recipient;
+
+    /**
+     * The type of Notification
+     *
+     * @var string $type
+     */
+    protected $type;
+
+    /**
+     * When is the Notification created
+     *
+     * @var DateTime $createdAt
+     */
+    protected $createdAt;
+
+    /**
+     * Notification constructor.
+     */
+    public function __construct()
+    {
+        $this->setCreatedAt(new DateTime());
+    }
+
+    /**
+     * Gets the Notification Subject
+     *
+     * @return string
+     */
+    public function getSubject(): string
+    {
+        return $this->subject;
+    }
+
+    /**
+     * Sets the Notification Subject
+     *
+     * @param string $subject
+     *
+     * @return $this
+     */
+    public function setSubject(string $subject)
+    {
+        $this->subject = $subject;
+
+        return $this;
+    }
+
+    /**
+     * Get the Notification Message
+     *
+     * @return string
+     */
+    public function getMessage(): string
+    {
+        return $this->message;
+    }
+
+    /**
+     * Set the Notification Message
+     *
+     * @param string $message
+     *
+     * @return $this
+     */
+    public function setMessage(string $message)
+    {
+        $this->message = $message;
+
+        return $this;
+    }
+
+    /**
+     * Returns the type of the Notification
+     *
+     * @return string
+     */
+    public function getType(): string
+    {
+        return $this->type;
+    }
+
+    /**
+     * Set the Notification type
+     *
+     * @param string $type
+     *
+     * @return $this
+     */
+    public function setType(string $type)
+    {
+        $this->type = $type;
+
+        return $this;
+    }
+
+    /**
+     * Get the recipient of this notification
+     *
+     * @return User
+     */
+    public function getRecipient(): User
+    {
+        return $this->recipient;
+    }
+
+    /**
+     * Set the recipient of this message
+     *
+     * @param User $recipient
+     *
+     * @return $this
+     */
+    public function setRecipient(User $recipient)
+    {
+        $this->recipient = $recipient;
+
+        return $this;
+    }
+
+    /**
+     * Get the creation date of this notification
+     *
+     * @return DateTime
+     */
+    public function getCreatedAt(): DateTime
+    {
+        return $this->createdAt;
+    }
+
+    /**
+     * Set the creation date of this notification
+     *
+     * @param DateTime $createdAt
+     *
+     * @return $this
+     */
+    public function setCreatedAt(DateTime $createdAt)
+    {
+        $this->createdAt = $createdAt;
+
+        return $this;
+    }
+}
--- a/site/src/Consumer/WebNotificationConsumer.php
+++ b/site/src/Consumer/WebNotificationConsumer.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace App\Consumer;
+
+use App\Util\Notification;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Class WebNotificationConsumer
+ *
+ * This class persists Notifications to the database
+ * and sends them out to the UI
+ *
+ * @author Aquib Baig <aquibbaig97@gmail.com>
+ */
+class WebNotificationConsumer extends AbstractNotificationConsumer
+{
+    /**
+     * WebNotificationConsumer constructor
+     *
+     * @param LoggerInterface $logger
+     */
+    public function __construct(LoggerInterface $logger)
+    {
+        parent::__construct($logger);
+    }
+
+    /**
+     * Handles Notifications that will be sent to UI
+     *
+     * @return bool
+     */
+    protected function handle(): bool
+    {
+        // TODO: Process the Notification($this->notification)
+        echo "Web Notification Consumer";
+
+        return true;
+    }
+}
+
+?>


imphil:
Thanks @aquibbaig for reducing the scope of this PR. It looks mostly good, but let's start in even smaller steps by only doing the RabbitMQ architectural stuff for now.

- Please remove the commits " create new Notification entity " and " Add Web Notifications " from the PR (use git rebase -i and remove the two commits, then force-push to this branch). Make sure you have your full work history in another branch.
- This needs a rebase first on top of the current master. So please do a 
  `git fetch origin` (or whatever the remote name of the upstream repository is) and then do a `git rebase origin/master` (again, replace the remote name)

aquibbaig:
Ah, I think the branch is fine. The **git log** returns 3 commits on top of master even. Also, rebasing is not adding any new changes.  We have 623 commits on master and 626 on this branch. So, it shouldn't be a problem

imphil:
It's fine indeed, no idea what problem github has here. The comment about removing all bug the rabbitmq commit still holds, however.

aquibbaig:
@imphil just force pushed the required changes.

site/app/config/packages/old_sound_rabbit_mq.yml
```diff
@@ -30,3 +33,13 @@ old_sound_rabbit_mq:
...
+        web_notification:
+            connection:       default
+            exchange_options: {name: 'send-notification', type: fanout}
+            queue_options:    {name: 'send-notification'}
+            callback:         web_notification_consumer
+        email_notification:
+            connection:       default
+            exchange_options: {name: 'send-notification', type: fanout}
+            queue_options:    {name: 'send-notification'}
+            callback:         email_notification_consumer
```
imphil: Isn't that callback the default and can be omitted?

site/app/config/services.yml
```diff
@@ -94,3 +97,6 @@ services:
     App\RepoCrawler\GitRepoCrawler:
         tags:
             - 'app.repo_crawler'
+
+    # Register Mgilet NotificationManager as a service to be used in WebNotificationConsumer
+    Mgilet\NotificationBundle\Manager\NotificationManager: '@mgilet.notification'
```
imphil: We discussed this before: This commit is titled " setup Rabbit MQ architecture with respective consumers ". the mgilet bundle is not needed here, and should not appear anywhere in this commit.

site/src/Consumer/AbstractNotificationConsumer.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * @var UserManagerInterface
+     */
+    protected $userManager;
+
+    /**
+     * AbstractNotificationConsumer constructor.
+     * @param UserManagerInterface $userManager
+     */
+    public function __construct(UserManagerInterface $userManager)
```
imphil: Why is the UserManagerInterface part of this interface?

site/src/Consumer/AbstractNotificationConsumer.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        $this->userManager = $userManager;
+    }
+
+    /**
+     * Decides if a notification should be handled by a particular Consumer
+     * or not by returning a boolean value
```
imphil: PHPDoc comments are structured like that:

```
/**
 * Summary line (a single line!)
 *
 * long description
 * 
 * @param
 * @return
 */

So please shorten the summary to a single line, like "Should the notification be handled by this class?"

File: site/src/Consumer/EmailNotificationConsumer.php [hunk 5]
```diff
@@ -0,0 +1,53 @@
...
+     * and persist it in the database
+     *
+     * @param AMQPMessage $msg
+     *
+     * @throws \Exception
+     */
+    public function execute(AMQPMessage $msg)
+    {
+        echo "\nEmail Notification Consumer\n";
+    }
```
imphil: Can you put the implementation of `execute()` into `AbstractNotificationConsumer`, and add another abstract method `handle()` which is then implemented in the child classes?

site/src/Util/NotificationHydrator.php
```diff
@@ -0,0 +1,119 @@
+<?php
+
+namespace App\Util;
+
+/**
+ * This is the data class for creating a notification
+ * which will be then sent out to the RabbitMQ queue
+ *
+ * @copyright Aquib Baig <aquibbaig97@gmail.com>
```
imphil: We optionally use `@author`, but not `@copyright`.

site/src/Util/NotificationHydrator.php
```diff
@@ -0,0 +1,119 @@
+<?php
+
+namespace App\Util;
+
+/**
+ * This is the data class for creating a notification
+ * which will be then sent out to the RabbitMQ queue
```
imphil: Please see my other comment about the structure of phpdoc comments.

site/src/Util/NotificationHydrator.php
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @copyright Aquib Baig <aquibbaig97@gmail.com>
+ *
+ * Class NotificationHydrator
```
imphil: What's that doing here?

site/src/Util/NotificationHydrator.php
```diff
@@ -0,0 +1,119 @@
...
+ */
+class NotificationHydrator
```
imphil: Notification*Hydrator* is a strange name. This is just a plain old data class (POD), so how about `Notification`?

site/src/Util/NotificationHydrator.php
```diff
@@ -0,0 +1,119 @@
...
+{
+    /**
+     * @var string $subject
+     */
+    protected $subject;
+
+    /**
+     * @var string $link
+     */
+    protected $link;
```
imphil: What's that for?

site/src/Util/NotificationHydrator.php
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * @var string $type
+     */
+    protected $type;
+
+    /**
+     * @var int $userIdentifier
+     */
+    protected $userIdentifier;
```
imphil: Is this the sender or the recipient? And it should be of type `User`; you can serialize that into a user id when sending.

site/src/Util/NotificationHydrator.php
```diff
@@ -0,0 +1,119 @@
...
+
```
imphil: This class is missing a bit of meta data: send timestamp, author

site/src/Service/NotificationProducerService.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Class NotificationProducer
+ *
+ * This class publishes a Notification to the RabbitMQ queue
+ * which is then taken up by the Consumers
+ *
+ * @author Aquib Baig <aquibbaig97@gmail.com>
+ */
+class NotificationProducerService
```
imphil: This class seems to be modeled after QueueDispatcherService, which isn't actually a particularly good design and needs to be fixed now that we have more than one producer service.

Please remove this class, you can inject the auto-generated service (i.e. the one you'd pass here as $notificationProducer) directly into the class method/class that needs to produce messages. (ProjectController.php in this case). Please don't use the `ProducerInterface`, but the actual class -- all producers conform to this interface, and autowire cannot really figure out which service class is the right one to bind now that there are two options.

site/app/config/packages/old_sound_rabbit_mq.yml
```diff
@@ -30,3 +33,13 @@ old_sound_rabbit_mq:
             exchange_options: {name: 'update-project-info', type: fanout}
             queue_options:    {name: 'update-github-metadata'}
             callback:         update_github_metadata_consumer
+        web_notification:
+            connection:       default
+            exchange_options: {name: 'send-notification', type: fanout}
+            queue_options:    {name: 'send-notification'}
```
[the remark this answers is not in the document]
aquibbaig: The message currently goes through the following areas
- An Exchange first
- Queue next
- Both the consumers where `shouldHandle()` checks if it will be handled by the one it is in

But I guess we want a structure where it goes through
- An Exchange first
- The Queue of the specific Consumer
- To the specific Consumer

site/app/config/services.yml
aquibbaig: This is a mistake I overlooked accidentally

site/app/config/packages/old_sound_rabbit_mq.yml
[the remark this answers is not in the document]
aquibbaig: I'll be working on this today itself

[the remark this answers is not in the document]
aquibbaig: So, I have come up with two promising architecture designs. I believe we need very small attention to detail here. I found there is another way here(it may be unnecessary).

# A Direct Exchange architecture
direct

**Advantages**
- Less message load in the buffer

# A fanout Exchange architecture
Fan

**Advantages**
- No routing key required as it broadcasts messages to both queues, simpler
- No logic required beforehand to generate a routing key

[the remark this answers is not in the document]
imphil: We want to go for the "fanout" architecture as inserting a message should be fast (it's in the synchronous page rendering path in many cases).

site/src/Util/NotificationHydrator.php
aquibbaig: Actually, Mgilet provides us with a `link` field by default which is nullable. So, we can either add/remove this field, later on, based on our usage

imphil: So let's do it the other way around: remove the link field in our `Notification(Hydrator)` class, as I don't see any use for it. If mgilet uses it that's fine, we can always set it to null.

site/app/config/packages/old_sound_rabbit_mq.yml
aquibbaig: Just checked that it expects a `callback` to be there else an error is thrown

site/src/Service/NotificationProducerService.php
aquibbaig: So, while publishing a message, do we need to get the producer just as 'this->get('@old_sound_rabbit_mq.notification_producer')' ?

imphil: You can inject the NotificationProducerService instead of using $this->get()

aquibbaig:
@imphil Just made the changes as discussed in yesterday's meeting. Added a few comments too, sorry for the mess the last time. I hope it's fine now. If not, please let me know

site/src/Consumer/AbstractNotificationConsumer.php
```diff
@@ -0,0 +1,70 @@
...
+     * @param AMQPMessage $msg
+     *
+     * @return bool|mixed
+     *
+     * @throws \Exception
+     */
+    public function execute(AMQPMessage $msg)
+    {
+        try {
+            $notification = unserialize($msg->body);
```
imphil: You can make $notification a member variable ($this->notification), and remove the argument from shouldHandle and handle()

site/src/Consumer/AbstractNotificationConsumer.php
```diff
@@ -0,0 +1,70 @@
...
+
+            return $this->handle($notification);
+        } catch (\Exception $e) {
+            throw $e;
+        }
```
imphil: Don't throw an exception here, it will abort the handler. Please look at the existing execute() methods in other code to see the possible return codes from this function.

site/src/Consumer/AbstractNotificationConsumer.php
```diff
@@ -0,0 +1,70 @@
...
+     * @return bool|mixed
+     *
+     * @throws \Exception
+     */
+    public function execute(AMQPMessage $msg)
+    {
+        try {
+            $notification = unserialize($msg->body);
+
+            return $this->handle($notification);
```
imphil: You need to call shouldHandle() as well.

site/src/Consumer/AbstractNotificationConsumer.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Should a Notification be handled by a particular Consumer?
+     *
+     * @param Notification $notification
+     *
+     * @return bool
+     */
+    abstract protected function shouldHandle(Notification $notification): bool;
```
imphil: You can always return true in the abstract implementation, it's a sane default.

site/src/Consumer/EmailNotificationConsumer.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace App\Consumer;
+
+use App\Util\Notification;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Class EmailNotificationConsumer
```
imphil: Can you make this summary line more descriptive, e.g. "Send out notifications over email"

site/src/Consumer/EmailNotificationConsumer.php
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @param Notification $notification
+     *
+     * @return bool
+     */
+    protected function shouldHandle(Notification $notification): bool
+    {
+        // TODO: Check if this Consumer should actually handle the Notification
+        return true;
+    }
```
imphil: You can remove this function if you implement it in the abstract class.

site/src/Consumer/EmailNotificationConsumer.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Handles an email Notification
+     *
+     * @param Notification $notification
+     *
+     * @return bool|mixed
```
imphil: where is the "mixed" coming from?

site/src/Consumer/EmailNotificationConsumer.php
```diff
@@ -0,0 +1,61 @@
...
+     */
+    protected function handle(Notification $notification)
+    {
+        if ($this->shouldHandle($notification)) {
```
imphil: This should go into the execute() function (see the comment there)

site/src/Util/Notification.php
```diff
@@ -0,0 +1,208 @@
...
+     * @var string $message
+     */
+    protected $message;
+
+    /**
+     * User associated with the Notification
+     *
+     * @var User $user
+     */
+    protected $user;
```
imphil: naming: user -> recipient

site/src/Util/Notification.php
```diff
@@ -0,0 +1,208 @@
...
+
+    /**
+     * Who issued the Notification
+     *
+     * @var string author
+     */
+    protected $author;
```
imphil: Hm, thinking about it, in almost all cases our messages come from the "system", so there's actually no sender/author. You can remove this field.

site/src/Util/Notification.php
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * When is the Notification created
+     *
+     * @var \DateTime $createdAt
+     */
+    protected $createdAt;
+
+    /**
+     * Notification constructor.
+     * @throws \Exception
```
imphil: does it?

site/app/config/packages/old_sound_rabbit_mq.yml
```diff
@@ -30,3 +33,14 @@ old_sound_rabbit_mq:
             exchange_options: {name: 'update-project-info', type: fanout}
             queue_options:    {name: 'update-github-metadata'}
             callback:         update_github_metadata_consumer
+    multiple_consumers:
```
imphil: Did you double-check that this leads to two queues?

aquibbaig: yep, it does

site/src/Util/Notification.php
aquibbaig: https://php.net/manual/en/datetime.construct.php

It emits an exception in case of error.

site/src/Consumer/EmailNotificationConsumer.php
aquibbaig: @imphil Please correct me out here. How do we implement logic that makes a particular Consumer handle a specific type of notification if we implement the `shouldHandle()` in the abstract class? Will we override that later in the consumers separately when we implement it?

site/src/Service/NotificationProducerService.php
aquibbaig: That's better

site/src/Consumer/EmailNotificationConsumer.php
aquibbaig: @imphil Otherwise I have pushed the requested changes

imphil: you override it if it should do something else than the default (which is implemented in the abstract class)

site/src/Consumer/AbstractNotificationConsumer.php
```diff
@@ -0,0 +1,89 @@
...
+     * @var Notification $notification
+     */
+    protected $notification;
+
+    /**
+     * AbstractNotificationConsumer constructor.
+     * @param LoggerInterface $logger
+     * @param Notification    $notification
+     */
+    public function __construct(LoggerInterface $logger, Notification $notification)
```
imphil: Please remove Notification from the constructor, it's not injected or passed there somehow: it's set in the execute() function.

site/src/Consumer/EmailNotificationConsumer.php
```diff
@@ -0,0 +1,41 @@
...
+ * Class EmailNotificationConsumer
+ *
+ * Sends out Notifications over email
+ *
+ * @author Aquib Baig <aquibbaig97@gmail.com>
+ */
+class EmailNotificationConsumer extends AbstractNotificationConsumer
+{
+    /**
+     * EmailNotificationConsumer constructor.
```
imphil: Add a empty line after this one.

site/src/Consumer/EmailNotificationConsumer.php
imphil: That comment needs resolving in a updated push.

site/src/Util/Notification.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Actual definition of the Notification
+     *
+     * @var string $message
+     */
+    protected $message;
+
+    /**
+     * User associated with the Notification
```
imphil: Please change this comment to match the variable (it's the recipient of the notification)

site/src/Util/Notification.php
```diff
@@ -0,0 +1,177 @@
...
+{
+    /**
+     * Subject of the Notification
+     *
+     * @var string $subject
+     */
+    protected $subject;
+
+    /**
+     * Actual definition of the Notification
```
imphil: Actual definition -> Notification message

site/src/Util/Notification.php
```diff
@@ -0,0 +1,177 @@
...
+     */
+    public function setType(string $type)
+    {
+        $this->type = $type;
+
+        return $this;
+    }
+
+    /**
+     * Returns User associated with a Notification
```
imphil: "Get the recipient of this notification"

site/src/Util/Notification.php
```diff
@@ -0,0 +1,177 @@
...
+     *
+     * @return User
+     */
+    public function getRecipient(): User
+    {
+        return $this->recipient;
+    }
+
+    /**
+     * Used to bind users to any Notification
```
imphil: "Set the recipient of this message"

site/src/Util/Notification.php
```diff
@@ -0,0 +1,177 @@
...
+     */
+    public function setRecipient(User $recipient)
+    {
+        $this->recipient = $recipient;
+
+        return $this;
+    }
+
+    /**
+     * Gets the timestamp of a Notification
```
imphil: "Get the creation date of this notification"

site/src/Util/Notification.php
```diff
@@ -0,0 +1,177 @@
...
+     *
+     * @return DateTime
+     */
+    public function getCreatedAt(): DateTime
+    {
+        return $this->createdAt;
+    }
+
+    /**
+     * Sets the timestamp of a Notification
```
imphil: do something equivalent to the get... description for this field.

aquibbaig: @imphil Yes. Just did that

imphil:
Thanks @aquibbaig, this now looks good! Merged & ready for new things!

PR #384, opened 2019-07-05
[GSoC 2019] Configure Web Notification Consumer

### LINKED ISSUE #374: [GSoC 2019] Store notifications in the database

aquibbaig:
Fixes #374

### DIFF
--- a/site/src/Consumer/WebNotificationConsumer.php
+++ b/site/src/Consumer/WebNotificationConsumer.php
@@ -2,8 +2,10 @@
 
 namespace App\Consumer;
 
-use App\Util\Notification;
+use App\Entity\Notification;
+use Doctrine\ORM\OptimisticLockException;
 use Psr\Log\LoggerInterface;
+use Mgilet\NotificationBundle\Manager\NotificationManager;
 
 /**
  * Class WebNotificationConsumer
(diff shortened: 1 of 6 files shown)


site/src/Entity/Notification.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    /**
+     * The recipient of the Notification
+     *
+     * @var User
+     *
+     * @ORM\Column(name="recipient", nullable=false)
+     *
+     */
+    protected $recipient;
```
imphil: You don't need that, the recipient is in the notifiable annotation in mgilet-notificationbundle.

site/src/Consumer/WebNotificationConsumer.php
```diff
@@ -15,27 +18,78 @@
...
-     * @param LoggerInterface $logger
+     * @param LoggerInterface     $logger
+     * @param NotificationManager $notificationManager
      */
-    public function __construct(LoggerInterface $logger)
-    {
+    public function __construct(
+        LoggerInterface $logger,
+        NotificationManager $notificationManager,
+        UserManagerInterface $userManager
```
imphil: this last argument is not documented in the phpdoc.

site/src/Consumer/WebNotificationConsumer.php
```diff
@@ -15,27 +18,78 @@
...
+        $notification = new Notification();
+
+        // Populate the Notification Entity
+        $notification->setMessage($this->notification->getMessage());
+        $notification->setType($this->notification->getType());
+        $notification->setSubject($this->notification->getSubject());
+        $notification->setDate($this->notification->getCreatedAt());
+        $notification->setRecipient($this->notification->getRecipient());
+
+        if ($this->shouldHandle()) {
```
imphil: Why do you call that method here? It has already been called once we enter handle().

site/src/Consumer/WebNotificationConsumer.php
```diff
@@ -15,27 +18,78 @@
...
 
-        return true;
+        $notification = new Notification();
+
+        // Populate the Notification Entity
+        $notification->setMessage($this->notification->getMessage());
+        $notification->setType($this->notification->getType());
+        $notification->setSubject($this->notification->getSubject());
+        $notification->setDate($this->notification->getCreatedAt());
+        $notification->setRecipient($this->notification->getRecipient());
```
imphil: The recipient is already given below in the notifiable part (the first parameter of addNotification()), so you don't need that here.

site/src/Consumer/WebNotificationConsumer.php
```diff
@@ -15,27 +18,78 @@
...
+
+        if ($this->shouldHandle()) {
+            // Persist a Notification
+            $this->notificationManager->addNotification(array($this->notification->getRecipient()), $notification, true);
```
imphil: Please keep this line around 100 characters.

The newer way of writing array() is just using []

site/src/Consumer/WebNotificationConsumer.php
aquibbaig: Agreed, But we are persisting the `$notification` variable which will not have a Recipient if we omit this line probably.

site/src/Consumer/WebNotificationConsumer.php
aquibbaig: `userManager` is not required in this Consumer, yet

site/src/Entity/Notification.php
aquibbaig: How can we fetch that while persisting the Notification for a specific user ?

aquibbaig: Cause not having a recipient while persisting crashes our consumer

site/src/Consumer/WebNotificationConsumer.php
imphil: The recipient is the "notifiable entity", you don't need a recipient field in addition to that.

site/src/Entity/Notification.php
imphil: You added an annotation to make the User entity a NotifableEntity; that's where the recipient of a message is stored in this case.

aquibbaig: That is correct, what was I even thinking!

aquibbaig:
@imphil Does the `handle()` method need more info on the logger and a `try->catch` ?

PR #385, opened 2019-07-08
Rendering web notifications to the UI and Notification Inbox

aquibbaig:
This is subject to further discussions regarding the design in the meeting today. I am adding that in the agenda. 
TODO: Add `Mark as seen` and/or `Remove Notification` buttons

### DIFF
--- a/site/app/Resources/views/base.html.twig
+++ b/site/app/Resources/views/base.html.twig
@@ -54,6 +54,26 @@
 
         <ul class="nav navbar-nav navbar-right">
           {% if is_granted("IS_AUTHENTICATED_REMEMBERED") %}
+            <li class="dropdown">
+              <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button"
+                 aria-haspopup="true" aria-expanded="false">
+                <i class="fa fa-bell"></i>
+                <span class="notification-count">{{ mgilet_notification_unseen_count(app.user) }}</span>
+                <span class="caret"></span>
+              </a>
+              <ul class="dropdown-menu list-group notification-list" role="menu">
+                <li class="list-group-item notification-list-title">
+                  <span class="item-left">Notifications</span>
+                  <span class="pull-right"><a href="#"><i class="fa fa-cog"></i></a></span>
+                </li>
+                <li class="divider"></li>
+                <li class="list-group-item notification-list-content">
+                  {{ mgilet_notification_render(app.user, { 'template': './notification/notification_list.html.twig', 'order': 'DESC'}) }}
+                </li>
+                <li class="divider"></li>
+                <li class="list-group-item notification-list-footer"><a class="text-center" href="">show all notifications</a></li>
+              </ul>
+            </li>
           <li class="dropdown">
             <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
               <i class="fa fa-plus"></i>
--- a/site/app/Resources/views/base.html.twig
+++ b/site/app/Resources/views/base.html.twig
@@ -71,7 +71,11 @@
                   {{ mgilet_notification_render(app.user, { 'template': './notification/notification_list.html.twig', 'order': 'DESC'}) }}
                 </li>
                 <li class="divider"></li>
-                <li class="list-group-item notification-list-footer"><a class="text-center" href="">show all notifications</a></li>
+                <li class="list-group-item notification-list-footer">
+                  <a class="text-center" href="{{ path('notification_inbox') }}">
+                    show all notifications
+                  </a>
+                </li>
               </ul>
             </li>
           <li class="dropdown">
--- a/site/app/Resources/views/base.html.twig
+++ b/site/app/Resources/views/base.html.twig
@@ -65,6 +65,7 @@
                 <li class="list-group-item notification-list-title">
                   <span class="item-left">Notifications</span>
                   <span class="pull-right"><a href="#"><i class="fa fa-cog"></i></a></span>
+                  <span class="pull-right"><button class="notification-markallseen">Mark all as seen</button></span>
                 </li>
                 <li class="divider"></li>
                 <li class="list-group-item notification-list-content">
--- a/site/app/Resources/views/notification/notification_list.html.twig
+++ b/site/app/Resources/views/notification/notification_list.html.twig
@@ -1,0 +1,29 @@
+{# Render notifications #}
+{% if mgilet_notification_unseen_count(app.user) != 0 %}
+  {% for notification in notificationList|filter(notification => notification.seen == false) -%}
+    {#Render a unseen notification #}
+    {% if (notification.seen == false) %}
+      <div class="unseen-notifications" id="notification-{{ notification.notification.id }}">
+        <p class="notification-subject">{{ notification.notification.subject }}
+          <span class="pull-right">
+            <button class="notification-markasseen"
+                    data-notification="{{ notification.notification.id }}">
+              <i class="fa fa-times markasseen-icon"></i>
+            </button>
+          </span>
+        </p>
+        <p class="notification-message">{{ notification.notification.message }}</p>
+        <p class="notification-date" data-livestamp="{{ notification.notification.date|date("c") }}">
+        </p>
+      </div>
+      <hr/>
+    {% endif %}
+  {% endfor %}
+{% else %}
+  <ul>
+    <li>You have seen all notifications!</li>
+  </ul>
+{% endif %}
+{% block javascripts %}
+  {{ encore_entry_script_tags('notification_list') }}
+{% endblock %}
--- a/site/app/Resources/views/user/notification_inbox.html.twig
+++ b/site/app/Resources/views/user/notification_inbox.html.twig
@@ -1,0 +1,55 @@
+{% extends 'layout.html.twig' %}
+
+{% block title %}
+  Notification Inbox
+{% endblock %}
+
+{% block pagepath %}Home &raquo; Notification Inbox{% endblock %}
+
+{% block stylesheets %}
+  {{ parent() }}
+  {{ encore_entry_link_tags('notification_inbox') }}
+{% endblock %}
+
+{% block content %}
+  <h1>Your notifications ({{ mgilet_notification_unseen_count(app.user) }})</h1>
+  <div class="inbox-header row">
+    <div class="text-right">
+      {# TODO: Update the URL to Notification Settings #}
+      <a class="inbox-settings" href="#">Notification Settings</a>
+    </div>
+  </div>
+  {# Render all notifications #}
+  <div class="all-notifications row">
+    {% for notification in notificationList %}
+      {# Single notification #}
+      <div class="inbox-notification">
+        <a href="{{ path('inbox_mark_seen', {notification: notification.id}) }}">
+          <ul class="list-group inbox-content">
+            <li class="list-group-item {{ (notification.seen) ? "inbox-seen" : "inbox-unseen" }}">
+              <span class="inbox-subject">
+                {{ notification.notification.subject }}
+              </span>
+              <form class="badge notification-remove"
+                  action="{{ path('remove_notification', {notification: notification.id}) }}">
+                <button class="btn btn-default btn-xs remove-button">
+                  <i class="fa fa-fw fa-trash"></i>
+                </button>
+              </form>
+              <p class="inbox-message">
+                {{ notification.notification.message|nl2br }}
+              </p>
+              <p class="inbox-notification-date" data-livestamp="{{ notification.notification.date|date("c") }}">
+              </p>
+            </li>
+          </ul>
+        </a>
+      </div>
+    {% endfor %}
+  </div>
+{% endblock %}
+
+{% block javascripts %}
+  {{ parent() }}
+  {{ encore_entry_script_tags('notification_inbox') }}
+{% endblock %}
--- a/site/assets/scss/app.scss
+++ b/site/assets/scss/app.scss
@@ -982,3 +982,45 @@ a.classifiation-tag:hover {
     }
   }
 }
+
+// Notification Dropdown
+.notification-list {
+  min-width: 330px;
+  background-color: #fff;
+  max-height: 280px;
+}
+.notification-list-content, .notification-list-title, .notification-list-footer {
+  border: 0;
+  padding: 0;
+}
+.notification-list-title > span {
+  padding: 0 10px 0 10px;
+}
+.notification-list-content {
+  min-height: 200px;
+  max-height: 200px;
+  overflow-y: scroll;
+}
+
+// Notification List
+.unseen-notifications {
+  list-style: none;
+  padding: 0 0 0 10px;
+}
+.unseen-notifications .notification-date {
+  font-size: 12px;
+  color: #B5B6B4;
+  padding:0;
+  margin:0;
+}
+.unseen-notifications .notification-subject {
+  font-weight: bold;
+}
+.unseen-notifications .notification-subject .notification-markasseen {
+  background-color: transparent;
+  border: 0;
+  padding: 0 10px 0 0;
+}
+.notification .markasseen-icon:hover {
+  color: #3eb0ac;
+}
--- a/site/assets/scss/app.scss
+++ b/site/assets/scss/app.scss
@@ -1024,3 +1024,11 @@ a.classifiation-tag:hover {
 .notification .markasseen-icon:hover {
   color: #3eb0ac;
 }
+.notification-markallseen {
+  background-color: transparent;
+  font-size: 10px;
+  border:0;
+}
+.notification-markallseen:hover {
+  color: #3eb0ac;
+}
--- a/site/assets/scss/notification_inbox.scss
+++ b/site/assets/scss/notification_inbox.scss
@@ -1,0 +1,53 @@
+// Notification Inbox
+.inbox-header {
+  border-bottom: 2px solid #e2e2e2;
+}
+.inbox-header .inbox-title {
+  font-weight: bold;
+  font-size: 16px;
+  padding-bottom: 10px;
+}
+.inbox-header a.inbox-settings {
+  text-decoration:none;
+  color: #3eb0ac;
+  font-weight: bold;
+}
+.inbox-header .inbox-settings:hover, .inbox-header a.inbox-settings:focus {
+  color: #3eb0ac;
+  font-weight: bold;
+}
+.all-notifications {
+  list-style: none;
+}
+.inbox-notification {
+  display: block;
+}
+.inbox-notification > a {
+  text-decoration: none;
+  color: black;
+}
+.all-notifications span.inbox-subject {
+  font-weight: bold;
+  padding-bottom: 5px;
+}
+.all-notifications .inbox-unseen, .all-notifications .inbox-seen {
+  padding: 0;
+}
+.all-notifications .inbox-unseen {
+  background-color: #eeeeee;
+}
+.all-notifications p.inbox-notification-date {
+  font-size: 10px;
+  display: inline-block;
+  color: #B5B6B4;
+}
+span.inbox-subject, p.inbox-message, p.inbox-notification-date {
+  padding-left: 5px;
+}
+ul.list-group.inbox-content {
+  margin-bottom: 0;
+}
+.notification-remove {
+  background-color: transparent;
+  border: 0;
+}
--- a/site/src/Controller/UserController.php
+++ b/site/src/Controller/UserController.php
@@ -3,6 +3,7 @@
 namespace App\Controller;
 
 use App\Entity\User;
+use App\Entity\Notification;
 use App\Form\Model\ResendConfirmationEmailRequest;
 use App\Form\Type\ResendConfirmationEmailRequestType;
 use App\Form\Type\UserProfileType;
@@ -13,6 +14,7 @@ use FOS\UserBundle\Model\UserInterface;
 use FOS\UserBundle\Model\UserManagerInterface;
 use FOS\UserBundle\Util\TokenGeneratorInterface;
 use Mgilet\NotificationBundle\Manager\NotificationManager;
+use Mgilet\NotificationBundle\NotifiableInterface;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
@@ -228,7 +230,7 @@ class UserController extends AbstractController
     }
 
     /**
-     * Mark a Notification as seen
+     * Mark notifications "seen" in the notification list
      *
      * @Route("/user/notification/seen", name="notification_mark_seen")
      *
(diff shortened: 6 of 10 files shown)


aquibbaig:
Screenshot 2019-07-08 at 3 11 47 PM

aquibbaig:
Progress: Added `Mark as seen` and `Remove` Button
Screenshot 2019-07-09 at 10 26 27 PM

aquibbaig:
Just had a check on a mobile viewport too

Screenshot 2019-07-09 at 10 29 01 PM

imphil:
Thanks for the UI proposal. I think we can simplify it a bit. How about this:

- Only show unseen notifications in the box.
- Remove the "Delete" functionality from the popup.
- Only keep the "seen" option.
- Remove the "eye" icon on the left, as now only unseen notifications are presented.
- Add time to the date.
- Make date/time less prominent, and use "relative dates" (we already include a library for that)

Furthermore, make sure that the notifications are "clickable", bringing the user to the "notification inbox". Do you have a screenshot for this page already?

aquibbaig:
@imphil I had not given much thought to the "Notification inbox" yet and that's the reason why you can see that it is not implemented yet. So, we would want a Notification inbox page where all the Notifications of the User reside along with these 2 handlers to mark them seen and remove them?

aquibbaig:
Also, the CSS in the first commit needs to be fixed. I will handle that!

imphil:
Sounds about right, but always try to think about simplifying the user experience. For example, you can avoid a "mark as seen" button if you just mark a notification as seen by clicking on it -- that's pretty standard in any email program. And the delete button is rather rarely used, so make it sufficiently small.

aquibbaig:
Thanks, @imphil for your suggestions on this pull request. After the meeting, I switched to this UI, let me know if there is anything bugging you in the design.
Screenshot 2019-07-10 at 11 59 10 PM

imphil:
The UI looks good now, with a minor nit on the text at the bottom: "See all Notifications" -> "show all notifications"

imphil:
I'll have a look at the code early tomorrow (UK time)

aquibbaig:
@imphil Regarding routing inside the controllers, there are two things to design now
- Notification Inbox
- Notification Settings

These two things are user preferences in my opinion and should be routed to /user/notification/settings and /user/notification/inbox rather than normal ones. So do you prefer moving the routing logic to UserController instead?

aquibbaig:
Adding some css fixes right now to separate seen notifications from unseen!

aquibbaig:
Can't seem to do that at the moment cause I can't open any page rn due to
`algoliaException\Hosts Unreachable`, which is most likely a result of either bad internet or some issues with my algolia app that I had created. I'll look into it first thing tomorrow morning

imphil:
On routing: We cannot use `/notification`, as this is outside of our available namespace for site pages. `/user/notifications` or `/user/inbox` is fine for the inbox, and settings should be `/user/settings/notifications`, in line with all the other settings pages we have today.

aquibbaig:
Screenshot 2019-07-12 at 11 18 57 PM

# Initial Design for the Notification Inbox
- Unseen notifications have a dark bg, also have a mark as seen and a remove handlers
- Seen notifications only have a remove handler and have transparent bg

aquibbaig:
@imphil Just moved the Notification Inbox logic to UserController. Will attempt the settings page tomorrow! Can you please have a look at what's been done already here, when you're free?

aquibbaig:
SInce the cross icon clashes with the markasseen cross in the dropdown, let me just quicky change it to a trash icon to symbolise remove action

imphil:
First UI comments:

image

aquibbaig:
@imphil  Added that details

> First UI comments:
> 
> image

imphil:
Initial comments:

UI: Still no padding!
image


General:
- Notifications -> notifications wherever it appears in a sentence (it's a normal word!)


Further testing depends on https://github.com/librecores/librecores-web/issues/392, I'll look into it (how did you test so far?)

aquibbaig:
@imphil The padding was there. Just update the branch as such. About `Notifications->notifications` is done now.

aquibbaig:
@imphil  For now, the command would create a Notification.
You can test it by running the consumer itself after that.
`php bin/console rabbitmq:multiple-consumer notification` is the command I guess

site/app/Resources/views/notification/notification_inbox.html.twig
```diff
@@ -0,0 +1,77 @@
...
+        {% if (mgilet_notification_count(app.user)) %}
+            {% for notification in notificationList %}
+                {% if (notification.seen) == true %}
+                    {# Render Seen Notification #}
+                    <li class="inbox-seen">
+                        <div class="row">
+                            <div class="col-sm-8 col-lg-8 col-xs-8 col-sm-8 inbox-message">
+                                {{ notification.notification.message }}
+                            </div>
+                            <button class="col-xs-4 col-md-4 col-sm-4 text-right notification-remove"
```
imphil: That column is way too large with 4/12 of the screen. You need a fixed-size column for the remove icon.

site/app/Resources/views/notification/notification_inbox.html.twig
```diff
@@ -0,0 +1,77 @@
...
+        <div class="col-xs-6 col-md-6 col-sm-6 inbox-title">
+            Your Notifications ({{ mgilet_notification_count(app.user) }})
+        </div>
+        <div class="col-xs-6 col-md-6 col-sm-6 text-right">
+            <a class="inbox-settings" href="/notification/settings">Notification Settings</a>
+        </div>
+    </div>
+    <ul class="all-notifications">
+        {# If there are any notifications #}
+        {% if (mgilet_notification_count(app.user)) %}
```
imphil: notificationList should be empty in that case, so you don't need to add the special check.

site/app/Resources/views/notification/notification_inbox.html.twig
```diff
@@ -0,0 +1,77 @@
...
+            {% for notification in notificationList %}
+                {% if (notification.seen) == true %}
```
imphil: Don't duplicate the whole block for seen/unseen. Instead, assign a single CSS class depending on the notification.seen property, and use CSS for further styling.

site/app/Resources/views/notification/notification_inbox.html.twig
```diff
@@ -0,0 +1,77 @@
...
+                    {# Render Seen Notification #}
+                    <li class="inbox-seen">
+                        <div class="row">
+                            <div class="col-sm-8 col-lg-8 col-xs-8 col-sm-8 inbox-message">
+                                {{ notification.notification.message }}
```
imphil: You need to add escaping for HTML special characters, and make sure that newlines get converted into HTML newlines.

site/app/Resources/views/notification/notification_inbox.html.twig
```diff
@@ -0,0 +1,77 @@
...
+        </div>
+    </div>
+    <ul class="all-notifications">
+        {# If there are any notifications #}
+        {% if (mgilet_notification_count(app.user)) %}
+            {% for notification in notificationList %}
+                {% if (notification.seen) == true %}
+                    {# Render Seen Notification #}
+                    <li class="inbox-seen">
+                        <div class="row">
```
imphil: Either use a HTML list, or Bootstrap rows, but not both.

site/app/Resources/views/notification/notification_inbox.html.twig
```diff
@@ -0,0 +1,77 @@
...
+            {% endfor %}
+        {% else %}
+            <li>No Notifications for you, yet!</li>
+        {% endif %}
+    </ul>
+{% endblock %}
+
+{% block javascripts %}
+    {{ parent() }}
+    {{ encore_entry_script_tags('notification') }}
```
imphil: Only include the JS for this page here (notification_inbox.js). All other JS should be included in the JS file itself (as import).

site/app/Resources/views/notification/notification_list.html.twig
```diff
@@ -0,0 +1,33 @@
+{# Render Notifications #}
+{% if mgilet_notification_unseen_count(app.user) != 0 %}
+    {% for notification in notificationList %}
+        {% if notification.seen == false %}
```
imphil: You should be able to combine the for and if by using `filter` (https://twig.symfony.com/doc/2.x/filters/filter.html).

site/assets/js/notification.js
```diff
@@ -0,0 +1,60 @@
...
+        const notificationId = $(this).attr("data-notification");
+        const notifiableId = $(this).attr("data-notifiable");
+
+        $.ajax({
+            url: "/notification/seen",
+            data: {'notification': notificationId, 'notifiable': notifiableId},
+            method: 'post',
+            success: function (response) {
+                if (response === 'seen') {
+                    console.log('Marked notification as seen...');
```
imphil: Remove such debug log statements. Leave them in for error conditions.

site/assets/js/notification.js
```diff
@@ -0,0 +1,60 @@
...
+                    window.location.reload();
+                }
+            },
+            error: function () {
+                console.log('Could not process that request')
+            }
+        })
+    });
+
+    // Remove a Notification on click
```
imphil: That code is unused, isn't it?

site/assets/scss/app.scss
```diff
@@ -982,3 +982,65 @@ a.classifiation-tag:hover {
...
+  background-color: #F5F5F5;
+}
+.notify-drop .drop-content::-webkit-scrollbar {
+  width: 8px;
+  background-color: #F5F5F5;
+}
+.notify-drop .drop-content::-webkit-scrollbar-thumb {
+  background-color: #ccc;
+}
+.notification-unseen {
```
imphil: For all these .notification-* classes: Use a single class like `notification` on the outer element (describing a single notification), and then only `unseen`, `seen`, etc. on the sub-elements. You can then do `.notification .unseen`, or nested blocks (it's SASS).

site/assets/scss/app.scss
```diff
@@ -982,3 +982,65 @@ a.classifiation-tag:hover {
...
+  padding: 10px 0 5px 0;
+  background-color: #eeeeee;
+}
+.notification-seen{
```
imphil: Style: space before {

site/src/Controller/NotificationController.php
```diff
@@ -0,0 +1,98 @@
...
+class NotificationController extends AbstractController
+{
+    /**
+     * @var NotificationManager
+     */
+    protected $notificationManager;
+
+    /**
+     * NotificationController constructor.
+     * @param NotificationManager $notificationManager
```
imphil: insert an empty line between the brief and `@param`.

site/webpack.config.js
```diff
@@ -64,6 +64,12 @@ Encore
     .addEntry('planet', [
         './assets/js/planet.js'
     ])
+    .addEntry('notification', [
```
imphil: Try to be consistent: the template is named "notification_list", how about using the same name here?

site/assets/js/notification.js
```diff
@@ -0,0 +1,60 @@
...
+        const notifiableId = $(this).attr("data-notifiable");
+
+        $.ajax({
+            url: "/notification/seen",
+            data: {'notification': notificationId, 'notifiable': notifiableId},
+            method: 'post',
+            success: function (response) {
+                if (response === 'seen') {
+                    console.log('Marked notification as seen...');
+                    window.location.reload();
```
imphil: *Either* you do client-side requests (then you need to hide/delete the items on the client side in addition to sending a request to the server), *or* you do normal form submissions (which reload the page, and then the list is freshly rendered with the remaining notifications). But don't do client-side JS and then reload the page.

site/src/Controller/NotificationController.php
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * @param Request $request
+     *
+     * @throws \Doctrine\ORM\OptimisticLockException
+     *
+     * @return JsonResponse
+     */
+    public function removeAction(Request $request)
+    {
+        if ($request->isXmlHttpRequest()) {
```
imphil: Return early, don't have long-ranging if statements.

```
if (!$request->isXmlHttpRequest()) {
  return new JsonResponse('Not an ajax request');
}
... continue here
```

site/src/Controller/NotificationController.php
```diff
@@ -0,0 +1,98 @@
...
+            $notification = $request->get('notification');
+            $notifiable = $request->get('notifiable');
+
+            // Deletes the link between a Notifiable and a Notification
+            $this->notificationManager->removeNotification(
+                [$this->notificationManager->getNotifiableInterface($this->notificationManager->getNotifiableEntityById($notifiable))],
+                $this->notificationManager->getNotification($notification),
+            );
```
imphil: Is this needed, isn't the deleteNotification() call below sufficient? If so, then the database schema is missing a couple foreign constraints.

imphil:
image

site/app/Resources/views/notification/notification_inbox.html.twig
```diff
@@ -0,0 +1,77 @@
...
+
+{% block stylesheets %}
+    {{ parent() }}
+    {{ encore_entry_link_tags('notification_inbox') }}
+{% endblock %}
+
+{% block content %}
+    <h1>Notification Inbox</h1>
+    <div class="inbox-header row">
+        <div class="col-xs-6 col-md-6 col-sm-6 inbox-title">
```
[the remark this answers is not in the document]
aquibbaig: Yes, I did

site/assets/js/notification.js
aquibbaig: I did a client side request and hid the notification. However any insight on how to update the notification count on the side of the notification bell icon (which gets updated on reload)

site/assets/js/notification.js
aquibbaig: Unfortunately, yes. Just removed it

site/assets/js/notification.js
aquibbaig: - Also, for form submits, we need to send a response to redirect the user to the page he was in currently when he clicks on the button, which we can't know as the dropdown is present in all pages. 
- Client-side request will not change the notification count immediately for which the user needs to refresh the page

site/src/Controller/NotificationController.php
aquibbaig: No, it doesn't. removed it

site/assets/js/notification.js
aquibbaig: @imphil I fixed the notification dropdown part and have pushed the changes.
I think since notification inbox is a separate page we can use form submit here

site/src/Controller/NotificationController.php
aquibbaig: Well, yes!
We need to add this and the foreign constraints are provided by the bundle, so no need to worry about that too!

aquibbaig:
@imphil I have switched to form actions in case of the notification inbox whereas ajax works perfectly fine for the notification-list. I have also modified the NotificationController to handle those request.

site/src/Controller/NotificationController.php
```diff
@@ -0,0 +1,131 @@
...
+     * @throws \Doctrine\ORM\EntityNotFoundException
+     * @throws \Exception
+     *
+     * @return Response
+     */
+    public function inboxMarkSeenAction($notification, $notifiable)
+    {
+        $this->notificationManager->markAsSeen(
+            $this->notificationManager->getNotifiableInterface($this->notificationManager->getNotifiableEntityById($notifiable)),
+            $this->notificationManager->getNotification($notification),
```
imphil: You can store the two parameters to the markAsSeen function in a variable to keep the line shorter and more readable. And re-use this variable later when getting the list.

site/src/Controller/NotificationController.php
```diff
@@ -0,0 +1,131 @@
...
+     */
+    public function __construct(NotificationManager $notificationManager)
+    {
+        $this->notificationManager = $notificationManager;
+    }
+
+    /**
+     * Mark a Notification as seen
+     *
+     * @Route("/notification/seen", name="notification_mark_seen")
```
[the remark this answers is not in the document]
aquibbaig: Will Notification Inbox be a part of User Controller?
SInce routes will now be changed to /user/notification
If that is the case, are we going to move the logic to userController?

aquibbaig:
@imphil Actually, clicking on a notification means that it has some content that the user will be redirected to. For example, in Facebook it implies a post, In gmail it implies entire messages. So, I guess putting a mark as seen and remove icons to do the same is fine. What do you think?

aquibbaig:
@imphil The ajax requests in the notification list return the count variable to the template. So, we handle the entire component after the ajax request has finished successfully. However in the notification inbox, we use forms to actually refresh the page which reflects changes in the notification list above and the list of notifications. It redirects to the notificatio inbox after the request. If we use the same routes(as in ajax request), we will have to handle two components(list and inbox) with js.

site/src/Controller/NotificationController.php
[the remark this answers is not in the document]
imphil: That would make sense, yes.

imphil:
>  So, I guess putting a mark as seen and remove icons to do the same is fine.

I'm not a huge fan of having two buttons there. Maybe another option would be to remove the "delete" button and only leave the mark as seen button?

site/app/Resources/views/base.html.twig
```diff
@@ -54,6 +54,29 @@
...
+            <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
+              <i class="fa fa-bell"></i>
+              <span class="notification-count">{{ mgilet_notification_unseen_count(app.user) }}</span>
+              <span class="sr-only"></span>
+              <span class="caret"></span>
+            </a>
+            <ul class="notification-menu dropdown-menu notify-drop">
+              <div class="notify-drop-title">
+                <div class="row">
+                  <div class="col-md-6 col-sm-6 col-xs-6 notification-title">notifications</div>
```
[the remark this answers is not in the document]
aquibbaig: @imphil I have no idea why GitHub is showing this. But this is already fixed with a rebase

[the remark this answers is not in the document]
aquibbaig: I'm fixing this in the next commit

aquibbaig:
Screenshot 2019-07-30 at 3 38 20 PM

@imphil Just tested with long texts such as this. Also now clicking on a notification marks it seen

site/app/Resources/views/notification/notification_inbox.html.twig
```diff
@@ -0,0 +1,62 @@
...
+                    <li class="inbox-subject">
+                        {{ notification.notification.subject }}
+                        {# Remove button #}
+                        <form class="pull-right" action="{{ path('remove_notification', {notification: notification.id, notifiable: notification.notifiableEntity.id}) }}">
+                            <button class="btn btn-default btn-xs"
+                                    data-notification="{{ notification.notification.id }}"
+                                    data-notifiable="{{ notification.notifiableEntity.id }}">
+                                <i class="fa fa-fw fa-times markasseen-icon"></i>
+                            </button>
+                        </form>
```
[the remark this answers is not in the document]
aquibbaig: @imphil This form just contains one button, Do we still need to make it using form builder?

site/src/Controller/NotificationController.php
```diff
@@ -0,0 +1,131 @@
...
+     * @throws \Doctrine\ORM\NonUniqueResultException
+     * @throws \Doctrine\ORM\EntityNotFoundException
+     * @throws \Exception
+     *
+     * @return JsonResponse
+     */
+    public function markAsSeenAction(Request $request)
+    {
+        $notification = $request->get('notification');
+        $notifiable = $request->get('notifiable');
```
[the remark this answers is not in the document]
aquibbaig: @imphil Now the logic is moved to userController where we fetch the user in the controller itself rather than sending it with the request. We get the notifiable accordingly there. Please have a check is the current approach alright. I have tested and it seems to work well

site/app/Resources/views/notification/notification_inbox.html.twig
```diff
@@ -0,0 +1,62 @@
...
+                    </li>
+                    <li class="inbox-message">
+                        {{ notification.notification.message|nl2br }}
+                        {# Mark as seen button #}
```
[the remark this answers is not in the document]
aquibbaig: Now, the user can click on a notification to mark it seen

agathver:
IMO, it would be to leave out the border around the buttons. The buttons to clear, or mark as read are not primary actions and hence do not need to be highlighted as such

site/app/Resources/views/user/notification_inbox.html.twig
```diff
@@ -0,0 +1,54 @@
...
+{% block content %}
+    <h1>Your notifications ({{ mgilet_notification_unseen_count(app.user) }})</h1>
+    <div class="inbox-header row">
+        <div class="text-right">
+            {# TODO: Update the URL to Notification Settings #}
+            <a class="inbox-settings" href="#">Notification Settings</a>
+        </div>
+    </div>
+    {# Render all notifications #}
+    <div class="all-notifications list-group">
```
imphil: - Remove the list-group here (use it on ul only)
- Add a "row" attribute to have the same padding as the previous div.

site/assets/scss/app.scss
```diff
@@ -982,3 +982,62 @@ a.classifiation-tag:hover {
...
+}
+.notification-markasseen {
+  background-color: transparent;
+  border: 0;
+}
+.notification-remove {
+  background-color: transparent;
+  border: 0;
+}
+.notification .markasseenIcon:hover {
```
imphil: We don't use lowerCamelCase in CSS classes, we use dashes. So this should be .markasseen-icon

site/assets/scss/notification_inbox.scss
```diff
@@ -0,0 +1,49 @@
...
+}
+.all-notifications p.inbox-notification-date {
+  font-size: 10px;
+  display: inline-block;
+  color: #B5B6B4;
+}
+span.inbox-subject, p.inbox-message, p.inbox-notification-date {
+  padding-left: 5px;
+}
+ul.list-group {
```
imphil: Don't change the behavior for all ul.list-group elements.

site/src/Controller/UserController.php
```diff
@@ -223,4 +228,143 @@ public function resendConfirmationEmailAction(
...
+    }
+
+    /**
+     * View all notifications in the user's inbox
+     *
+     * @Route("/user/notification/inbox", name="notification_inbox")
+     *
+     * @param NotificationManager $notificationManager
+     *
+     * @return string;
```
imphil: remove the ;

site/src/Controller/UserController.php
```diff
@@ -223,4 +228,143 @@ public function resendConfirmationEmailAction(
...
+    {
+        $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');
+
+        $user = $this->getUser();
+
+        $notifiableEntity = $notificationManager->getNotifiableEntity($user);
+        $notifiable = $notificationManager->getNotifiableInterface($notificationManager->getNotifiableEntityById($notifiableEntity));
+
+
+        // Deletes the link between notification and notifiable
```
imphil: I think I asked this before: Why don't take foreign constraints care of deleting all parts of a notification? If they do (as they should), you should be OK with only one remove* function call. Having only one such call is really helpful: now your remove becomes a single transaction and the database takes care that no intermediate state becomes visible (e.g. a notification being half-deleted).

aquibbaig: @imphil We will resolve this discuss in the meeting today. It's a bit tricky one!

aquibbaig: Added a comment in the code as discussed

aquibbaig:
@imphil just edited the commits w.r.t. yesterday's discussion. Let me know what you think

site/app/Resources/views/notification/notification_list.html.twig
```diff
@@ -0,0 +1,32 @@
+{# Render notifications #}
+{% if mgilet_notification_unseen_count(app.user) != 0 %}
+  {% for notification in notificationList|filter(notification => notification.seen == false) -%}
+    {#Render a unseen notification #}
+    {% if (notification.seen == false) %}
+      <ul class="unseen-notifications" id="notification-{{ notification.notification.id }}">
```
imphil: Please don't misuse ul/li as generic block element here and in the base.html.twig. Use lists when there are actually things to list (like a list of notifications), but use <div> if you only have blocks next to each other.

site/app/Resources/views/user/notification_inbox.html.twig
```diff
@@ -0,0 +1,54 @@
+{% extends 'layout.html.twig' %}
+
+{% block title %}
+    Notification Inbox
```
imphil: Please indent this file by two spaces and double-check all files this time please.

site/src/Controller/UserController.php
```diff
@@ -223,4 +228,146 @@ public function resendConfirmationEmailAction(
...
+        $notificationManager->removeNotification(
+            [$notifiable],
+            $notificationManager->getNotification($notification),
+            true
+        );
+
+        // Remove the Notification from the database
+        // TODO: Let foreign constraints take care of removing notifications, see #400
+        // $notificationManager->deleteNotification($notificationManager->getNotification($notification), true);
```
imphil: Wait: you commented this out, so it works without this line?

aquibbaig: Yup, it does. As I said yesterday removeNotification() removes link between notifiableEntity and notifiableNotification, so if that is removed the user can't see it anymore

aquibbaig: @imphil The deleteNotification() method is for removing the notification from the database itself

site/app/Resources/views/user/notification_inbox.html.twig
aquibbaig: Sure!

site/src/Controller/UserController.php
aquibbaig: It is a workaround probably

site/app/Resources/views/notification/notification_list.html.twig
aquibbaig: @imphil I had some checks on how others do it and this seems to be a general case(for bootstrap > 3.0). We just have a twig renderer that contains <li> items. <li> can have a nested <ul> and then <li> inside as per documentation. Please have a small look at this [Bootstrap snippet](https://bootsnipp.com/snippets/1e8Ar). So, are we misusing it still?

site/src/Controller/UserController.php
imphil: So the notification is still half-present in the database? That's not what we want, the notification should be fully gone. Sothe deleteNotification() part is necessary.

site/app/Resources/views/notification/notification_list.html.twig
imphil: ul/li is semantic markup for a list. A list of notifications is a list. A list of elements in a navigation bar is a list. Use ul/li there. But the code you have here is not a list: it's just different paragraphs: a subject, a message, a date, a divider. That's not a list, and therefore it should not use ul/li.

site/src/Controller/UserController.php
aquibbaig: @imphil Exactly!

site/app/Resources/views/notification/notification_list.html.twig
aquibbaig: @imphil Sounds logical indeed, I'm using `<span>` and `<div>`s

imphil:
Thanks @aquibbaig, I've now merged your PR!

aquibbaig:
Final Preview
Screenshot 2019-08-20 at 10 12 24 PM

PR #401, opened 2019-08-04
Create a notification

aquibbaig:
We have already bound the `notificationProducer` to rabbit-mq producer in services, so no need to do that now.


imphil:
can you modify CreateNotificationCommand.php to use this new service?

aquibbaig:
> can you modify CreateNotificationCommand.php to use this new service?

Yeah sure. This will be added in this pull request

PR #404, opened 2019-08-17
tests for web notification consumer

aquibbaig:
Tried writing some tests for the web notification consumer. @imphil @agathver Please suggest if I have missed something on this.

Since I found that protected/private methods should not be tested, I did some refactoring to the existing code itself to test out the `handle()` function

### DIFF
--- a/site/tests/RepoCrawler/WebNotificationConsumerTest.php
+++ b/site/tests/RepoCrawler/WebNotificationConsumerTest.php
@@ -1,0 +1,184 @@
+<?php
+
+namespace App\Tests\RepoCrawler;
+
+use App\Consumer\WebNotificationConsumer;
+use App\Entity\User;
+use App\Util\Notification;
+use Mgilet\NotificationBundle\Manager\NotificationManager;
+use PhpAmqpLib\Message\AMQPMessage;
+use PHPUnit\Framework\TestCase;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Tests for the Web Notification Consumer
+ *
+ * @author Aquib Baig <aquibbaig97@gmail.com>
+ */
+class WebNotificationConsumerTest extends TestCase
+{
+    protected $notification;
+
+    /**
+     * Called before every test
+     */
+    public function setUp()
+    {
+        $this->notification = new Notification();
+        $this->notification->setMessage('Your project was added to LibreCores')
+            ->setSubject('New Project')
+            ->setType('new_project');
+        $user = new User();
+        $user->setUsername('test');
+        $this->notification->setRecipient($user);
+
+        // Assert if fields are set properly
+        $this->assertEquals('Your project was added to LibreCores', $this->notification->getMessage());
+        $this->assertEquals('New Project', $this->notification->getSubject());
+        $this->assertEquals('new_project', $this->notification->getType());
+        $this->assertEquals('test', $this->notification->getRecipient()->getUsername());
+    }
+
+    /**
+     * Called after every test
+     */
+    public function tearDown()
+    {
+        $this->notification = null;
+    }
+
+    /**
+     * @test
+     *
+     * Test if notification of a specific type are handled
+     */
+    public function testIfMentionedTypesOfNotificationsAreHandled()
+    {
+        $mockLogger = $this->createMock(LoggerInterface::class);
+        $mockNotificationManager = $this->getMockBuilder(NotificationManager::class)
+            ->setMethods(['addNotification'])
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $webNotificationConsumer = new WebNotificationConsumer($mockLogger, $mockNotificationManager);
+        $webNotificationConsumer->setNotification($this->notification);
+
+        $msg = new AMQPMessage();
+        $msg->body = serialize($this->notification);
+
+        $mockWebNotificationConsumer = $this->getMockBuilder(WebNotificationConsumer::class)
+            ->setMethods([ 'handle'])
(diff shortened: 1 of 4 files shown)


site/tests/RepoCrawler/WebNotificationConsumerTest.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->assertEquals('new_project', $this->notification->getType());
+        $this->assertEquals('test', $this->notification->getRecipient()->getUsername());
+    }
+
+    /**
+     * @test
+     *
+     * Tests if a notification is added to database
+     */
+    public function test_if_notification_is_added()
```
imphil: use lowerCamelCase() for function names.

site/tests/RepoCrawler/WebNotificationConsumerTest.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $mockLogger = $this->createMock(LoggerInterface::class);
+        $mockNotificationManager = $this->getMockBuilder(NotificationManager::class)
+            ->setMethods(['addNotification'])
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $webNotificationConsumer = new WebNotificationConsumer($mockLogger, $mockNotificationManager);
+        $webNotificationConsumer->setNotification($this->notification);
```
imphil: That doesn't test what is really happening. What you should do instead is run the execute() function with a serialized notification object.

site/tests/RepoCrawler/WebNotificationConsumerTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+        // Test if handle() doesn't requeue notifications in RabbitMQ
+        $this->assertEquals(true, $webNotificationConsumer->handle());
```
imphil: Don't put the function call into the assert

site/tests/RepoCrawler/WebNotificationConsumerTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+        // Test if notification is added to database
+        $mockNotificationManager->expects($this->once())
+            ->method('addNotification');
+
+        $webNotificationConsumer->handle();
```
imphil: What's that for? You already called handle() before.

site/tests/RepoCrawler/WebNotificationConsumerTest.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+}
+
+?>
```
imphil: Leave that out.

site/tests/RepoCrawler/WebNotificationConsumerTest.php
```diff
@@ -0,0 +1,66 @@
...
+        // Test if handle() doesn't requeue notifications in RabbitMQ
+        $this->assertEquals(true, $webNotificationConsumer->handle());
+
+        // Test if notification is added to database
+        $mockNotificationManager->expects($this->once())
+            ->method('addNotification');
+
+        $webNotificationConsumer->handle();
+    }
+}
```
imphil: The test_if_notification_is_added() test is incomplete, it doesn't explore all possible paths the function can take.
- What happens if shouldHandle() returns true or false? How can you test both cases?
- In which cases returns handle() a 0 or non-zero return value? What can do you to test both cases?

aquibbaig: @imphil For testing both shouldHandle and handle, we need to set them as public first?

aquibbaig: Else we need to have a public method that calls them, what's the better option?

aquibbaig: @imphil I just made the possible tests, please have a look at it

site/tests/RepoCrawler/WebNotificationConsumerTest.php
```diff
@@ -0,0 +1,187 @@
...
+        $this->assertEquals('new_project', $this->notification->getType());
+        $this->assertEquals('test', $this->notification->getRecipient()->getUsername());
+    }
+
+    /**
+     * Called after every test
+     */
+    public function tearDown()
+    {
+        $this->notification = [];
```
imphil: Why do you set the notification variable to an array?

site/tests/RepoCrawler/WebNotificationConsumerTest.php
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * @test
+     *
+     * Test if notification of a specific
+     * type are handled
+     */
+    public function testIfMentionedTypesOfNotificationsAreHandled()
+    {
+
+
```
imphil: Remove empty lines here.

site/tests/RepoCrawler/WebNotificationConsumerTest.php
```diff
@@ -0,0 +1,187 @@
...
+    public function tearDown()
+    {
+        $this->notification = [];
+    }
+
+    /**
+     * @test
+     *
+     * Test if notification of a specific
+     * type are handled
```
imphil: make that one line

aquibbaig:
@imphil Fixed!

site/tests/RepoCrawler/WebNotificationConsumerTest.php
aquibbaig: I have set it to null now

PR #423, opened 2019-09-02
Get recipient from notification rather than UserManager

aquibbaig:


site/src/Consumer/EmailNotificationConsumer.php
```diff
@@ -82,9 +80,9 @@ protected function shouldHandle()
      */
     public function handle(): bool
     {
-        echo "Email Notification Consumer";
+        echo "Email Notification Consumer\n";
```
imphil: You can remove this echo statement.

site/src/Consumer/AbstractNotificationConsumer.php
```diff
@@ -72,6 +81,11 @@ public function setNotification(Notification $notification)
         $this->notification = $notification;
     }
 
+    public function setRecipient(User $user)
+    {
+        $this->recipient = $user;
+    }
+
```
imphil: You have this information already available in `$notification`, no need for a separate `$recipient` variable. (Actually, it's harmful to have a second copy of the same information, as the two could get out of sync.)

site/src/Consumer/EmailNotificationConsumer.php
```diff
@@ -59,16 +59,14 @@ public function __construct(
 
     protected function shouldHandle()
     {
-        $user = $this->userManager->findUserByUsername($this->notification->getRecipient()->getUsername());
-
-        $emailSubscription = $user->isSubscribedToEmailNotifications();
+        $emailSubscription = $this->recipient->isSubscribedToEmailNotifications();
```
imphil: Use `$this->notification->getRecipient()` instead.

aquibbaig:
Thanks @imphil for your review on this PR. I have updated the branch now as per your review.

site/src/Consumer/AbstractNotificationConsumer.php
aquibbaig: Yep

imphil:
Please have a closer look at the diff. There should be no changes in `AbstractNotificationConsumer.php`

aquibbaig:
@imphil I had forgotten to `git reset HEAD` the abstractNotificationConsumer file. Should be fine now!

Issue #402, opened 2019-08-08
Add documentation for SwiftMailer configuration in  dev


aquibbaig:
For sending out email notifications to users in dev, a specific configuration must be set in parameters.yml file, so we need to mention the same in our documentation

imphil:
Actually, the configuration should happen in `secrets/dev-vagrant.secrets.yml`. The way to modify this file is to copy it over from secrets.dist and then modify it. The file is commented, but adding a specific example for Gmail will be a good thing.

aquibbaig:
Okay. Should the `Sender's email address` be defined in this file?

aquibbaig:
> Actually, the configuration should happen in `secrets/dev-vagrant.secrets.yml`. The way to modify this file is to copy it over from secrets.dist and then modify it. The file is commented, but adding a specific example for Gmail will be a good thing.

There can be a better solution. We will use the current swiftmailer configuration as the default case but we will use another mailer `notification_mailer` for email notifications. That configuration will be optimal for everything I guess

aquibbaig:
@imphil I'm adding some changes to the email consumer. Please review when free

imphil:
Comment by @agathver:Also add null configuration as default to test spooling of emails.

PR #424, opened 2019-09-09
documentation for the notification-system

aquibbaig:
Initial commit aimed at solving #402


imphil:
Thanks! (I don't know what this one was lingering around here for so long, seems like I've forgotten about it. Sorry.)

PR #427, opened 2019-10-18
Send notification on join-organisation request

### LINKED ISSUE #411: Send notification on join organization request
We need to send a notification when a "join organization" request is issued. This notification should go out to all current org members.

aquibbaig:
Fixes #411

### DIFF
--- a/site/src/Controller/OrganizationController.php
+++ b/site/src/Controller/OrganizationController.php
@@ -5,10 +5,13 @@ namespace App\Controller;
 use App\Entity\Organization;
 use App\Entity\OrganizationMember;
 use App\Repository\OrganizationRepository;
+use App\Service\NotificationCreatorService;
+use App\Util\Notification;
 use Doctrine\ORM\NonUniqueResultException;
 use FOS\UserBundle\Model\UserManagerInterface;
 use App\Form\Type\OrganizationType;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -189,16 +192,21 @@ class OrganizationController extends AbstractController
     /**
      * Request to join an organization
      *
-     * @param string                 $organizationName
-     *
+     * @param string $organizationName
      * @param OrganizationRepository $repository
+     * @param NotificationCreatorService $notificationCreatorService
+     * @param ContainerInterface $container
      *
      * @return Response
      *
      * @throws NonUniqueResultException
      */
-    public function joinAction($organizationName, OrganizationRepository $repository)
-    {
+    public function joinAction(
+        $organizationName,
+        OrganizationRepository $repository,
+        NotificationCreatorService $notificationCreatorService,
+        ContainerInterface   $container
+    ) {
         $o = $repository->findOneByName($organizationName);
 
         if (!$o) {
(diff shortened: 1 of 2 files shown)


site/src/Controller/OrganizationController.php
```diff
@@ -215,6 +220,23 @@ public function joinAction($organizationName, OrganizationRepository $repository
         $member->setUser($user);
         $member->setPermission(OrganizationMember::PERMISSION_REQUEST);
         $em = $this->getDoctrine()->getManager();
+        // Find Organisation members
+        $orgMembers = $repository->findOneByName($organizationName)->getMemberUsers();
```
imphil: This should be limited to org admins, normal users cannot accept/reject a request to join.

site/src/Controller/OrganizationController.php
```diff
@@ -215,6 +220,23 @@ public function joinAction($organizationName, OrganizationRepository $repository
...
+        // Queue notifications to each member
+        $notification = new Notification();
+        foreach ($orgMembers as $recipient) {
+            // Prevent mailing the currentUser
+            if ($recipient !== $this->getUser()) {
```
imphil: Wouldn't this mean that the user requesting to join an org is already part of it? I think this check can be removed.

site/src/Controller/OrganizationController.php
```diff
@@ -215,6 +220,23 @@ public function joinAction($organizationName, OrganizationRepository $repository
...
+                $notification->setType("organisation_join")
+                    ->setSubject("Join organisation request")
+                    ->setRecipient($recipient)
+                    ->setMessage(
+                        $member->getUser()->getUsername()." has requested to join your organisation.
+                    Kindly go to the Librecores website to accept/reject the request."
```
imphil: Please include the organization name, and a link to the org admin page (or directly link to the accept/reject actions). And please fix the indentation.

aquibbaig:
Thanks, @imphil for the review. I will look into the following changes.

aquibbaig:
@imphil Does these settings make sense? I believe the message can be improved further. I will test them out at once.

site/src/Controller/OrganizationController.php
```diff
@@ -215,6 +220,20 @@ public function joinAction($organizationName, OrganizationRepository $repository
...
+        // Find Organisation members
+        $orgAdmin = $repository->findOneByName($organizationName)->getCreator();
+        // Queue notifications to the organisation admin
+        $notification = new Notification();
+        $notification->setType("organisation_join_request")
+            ->setSubject("Organisation join request")
+            ->setRecipient($orgAdmin)
+            ->setMessage(
+                $member->getUser()->getUsername()." has requested to join ".
+                $organizationName . ". Kindly go to https://www.librecores.org/"
```
imphil: Please use the router component to construct this URL, and ensure that the domain is not hardcoded but depends on the configuration (or in this case, on the request).

site/src/Controller/OrganizationController.php
```diff
@@ -215,6 +220,20 @@ public function joinAction($organizationName, OrganizationRepository $repository
         $member->setUser($user);
         $member->setPermission(OrganizationMember::PERMISSION_REQUEST);
         $em = $this->getDoctrine()->getManager();
+        // Find Organisation members
+        $orgAdmin = $repository->findOneByName($organizationName)->getCreator();
```
imphil: Please send this notification to all admins (can be more than one). They are OrganizationMember entities with the PERMISSION_ADMIN permission set.

aquibbaig:
Thanks @imphil on the follow up. I almost had forgot about that `url` parameter we created during last summer and you're completely right about using that. I have updated the PR. WIll test it out soon

site/src/Controller/OrganizationController.php
aquibbaig: +1

